Issue #4, opened 2016-12-21
Framework: enable CI testing


aprokop:
This issues is to track status of Jenkins CI testing.
- [x] Create a Docker container with Trilinos TPL similar to how DTK does it
- [x] Make sure tests pass in the container
- [x] Hook up Jenkins CI with Github
- [x] Configure Jenkins to launch CI on PR

aprokop:
Fixed in #5.

jwillenbring:
@aprokop How much would would it be to generalize what you did here to work for all of Trilinos?

aprokop:
@jwillenbring Not much. Essentially, just need to update the Dockerfile script to provide all TPLs required for testing, and create Jenkins configuration to run things. Though, I was not the person who set up the Jenkins installation from scratch, I just used the existing infrastructure of the [DTK](https://github.com/ORNL-CEES/DataTransferKit). But I know the person who set it up.

If you are interested, I can provide more info on how things work.

jwillenbring:
@aprokop I would be interested. How much time would you estimate you spent on this?

aprokop:
Around couple hours to set up the Docker, and a day for Jenkins (I think I got hit by some bug in Jenkins where copying an existing configuration is *not* the same as creating one from scratch, even though at the end they would have the exact same configuration).

aprokop:
Just testing Slack integration.

aprokop:
My colleague has accidentally deleted the Jenkins configuration for CI. I need to resurrect it now.

aprokop:
OK, redid the hooks and Jenkins configuration. Works on PR now.

aprokop:
We currently have only testing for PRs, but we do not test `develop` branch after merging. This should be done.

aprokop:
My first at CI on `develop` does not work. I have a Jenkins configuration, and a webhook installed, but the webhook is not triggered on the Github side with a merged PR.

aprokop:
Alright, I did not change anything, but my colleague did some changes to the Jenkins config, and now things seem to work. I see Github payload delivered to Jenkins, and build triggered. Will test on one more PR merge, and if things go right we should get a proper CI badge in README on `develop`.

aprokop:
Still, no clue about why Jenkins badge does not display properly. Talking to Damien did not help. Googling did not help either.

Thing searched:
- Aggressive Github caching
- Reverse proxies
- shields.io workaround
- ...

PR #5, opened 2016-12-27
ForTrilinos: enable Jenkins testing based on Docker

aprokop:
This is based on existing testing framework in
ORNL-CEES/DataTransferKit.


aprokop:
Fixes #4.

aprokop:
All the close/open above is debugging the hooks before I learned I can simply resend the payload to Jenkins through the webhooks configuration option.

Issue #8, opened 2017-02-03
Create a shortcut approach to using Trilinos


aprokop:
While the eventual goal is to expose a significant portion of Trilinos functionality to Fortran applications, it is a good idea to expose the potential applications to the benefits of Trilinos as early as possible.

The idea is to provide a minimal interface (not existing in Trilinos proper at this point) which would solve a linear systems given few Fortran arrays with data corresponding to matrix and vector, a communicator, and a parameter list. Something like this (on C++ side):
```c++
int solve_linear_system(
  MPI_Comm comm,
  const int * row_ptrs, int num_rows,
  const int * col_inds, int num_values 
  const double * vals, int num_values,
  const double * rhs, int size,
  double * sol, int size,
  const ParameterList* list);
```
Not all sizes are necessary, but it makes it easier to convert from Fortran.

On the C++ side we wrap the data in the proper Trilinos classes (and constructing missing ones like `Map`), and pass it to Stratimikos.

sethrj:
Ross used static RCPs to allow this kind of thing to be broken into multiple function calls; maybe we'd want to do something similar to allow a little more flexibility?

aprokop:
We certainly could, but that would also require storing temporary handles on the Fortran side between code. Having it in one function would bypass that.

sethrj:
not really, he only needed handles because he used multiple instances; for simple trilinos we'd only need a single instance; and all the function calls would be module-level functions

sethrj:
I envision an interface like:
```
void init();
void init(MPI_COMM comm);
void build_crs(int size, int* i, int* j, double* v);
void set_params(RCP<ParameterList>& p);
void solve(int size, const double* b, double* x);
void finalize();
```
and on the backend have
```
RCP<Comm> comm;
RCP<Solver> solver;
```
and so forth that get built during `init` and reset in `finalize`

aprokop:
Here is a modified version:
```C++
void init();
void init(MPI_COMM comm);
void setup_matrix(int num_rows, const int* i, int num_nnz, const int* j, const double* v);
void set_params(const RCP<ParameterList>& p);
void setup_solver();
void solve(int size, const double* b, double* x);
void finalize();
```
The main difference is adding a second size for nnz in the matrix construction, and separating the solver setup from the solve. This should allow a more clear idea about when a solver is constructed, and when a system is solved.

sethrj:
[...]

PR #14, opened 2017-02-22
ForTrilinos: adding simple interface

aprokop:
Partial #8


aprokop:
Updated PR. Few things are still missing, but getting close. People can start looking at the interface.

aprokop:
@sethrj I had to do a small tweak to `ParameterList` module in Fortran by making `ptr` public. Otherwise, how do we pass ParameterList in other modules inside Fortran proxies? Nota bene: I most likely miss something.

maherou:
I like the pattern of the simple code.  As I read it, I notice that some of the design details can go a variety of ways.  Might it be worth vetting this code with some other Fortran app developers.  I am wondering if we can ID a few savvy Fortran users in the broader community to critique the design.

aprokop:
@maherou This is definitely not the final design, and we will undoubtedly hit some real world limitations with it. What are the things that caught your eye?
And I agree that we should talk more to other folks.

I see some few obvious limitations of the current design:
1.  Its inability to work with multiple matrices simultaneously
2. Its strict requirement to deal with GIDs

maherou:
The detail that I noticed is your item 1.  There is a bit too much magic for my tastes to see that the matrix is constructed but there is no object I can point to.  At the same time, I don't know the sensibilities of modern Fortran programmers, so I don't want to dismiss this approach without getting some input.

To me, having an object that I pass into the solver feels right, in addition to being able to have more than one linear problem, and similarly more than one solver instantiation available at one time.

aprokop:
Ready for review.

aprokop:
Updated PR to make `SimpleInterface` non-singleton, allowing solution of multiple systems simultaneously.

aprokop:
Actually, it just seems to run, but also is buggy which needs fixing:

- Indices treatment: 1-based vs 0-based
- Need to add test to check solution

aprokop:
Updated PR. Now I'm actually checking that the solutions are correct.

aprokop:
Added inversion of control for linear operators. The task itself is tracked in #22. Now, the only thing that seems remaining is to provide IoC for preconditioners. This may take a bit more time.

aprokop:
Hmm, the test fails though it runs fine for me. But, of course, I can reproduce it thanks to our container-based testing. I like it better and better every time I need to debug issues like this.

Anyways, I need to figure out what is the difference in Trilinos configuration between two.

aprokop:
OK. Ready for review.

youngmit:
I just looked through the standard, and it makes no guarantees about the order of evaluation of operands in logical statements. Therefore, it isn't safe to do something like:

    function bar(array) result(ret)
        real, intent(in) :: array(*)
        logical :: ret
        !...
    end subroutine
    real,pointer :: foo(:)
    !...
    if(associated(foo), bar(foo)) then
        !...
    end if

That said the test that raised the question is still fine, this is just for future reference.

aprokop:
@sethrj Oh yes, you are right, I did want to change it to make MPI optional.

Yes, you are right about Belos and Ifpack2 being required for the provided input deck.

What changes are you referring to (SWIG automation)? I'd like to keep ierr as a parameter, but I could regenerate and then add it.

sethrj:
The changes are in the `simple_seth` branch I pushed yesterday (you'll need the latest sethrj/swig/fortran branch).

Sure, ierr is more fortran-like, but unless you explicitly add it to your C interface, it's not something sustainable (since SWIG isn't able to automate adding an 'ierr' parameter). Will and I have a discussion about exception support [here](https://github.com/sethrj/swig/issues/10) and it would be good to decide on how to do it, and implement it, before we start hand-rolling a thousand different functions and their arguments.

aprokop:
@sethrj I've updated the PR to address your comments:
- Added optional test dependencies
- Made MPI optional

aprokop:
OK. I'm merging now, and then will incorporate bits and pieces from `simple_seth` branch.

PR #20, opened 2017-02-24
ForTrilinos: more doc update

aprokop:
- Added ReadTheDocs badge
- Added Doc reference in README
- Added some notes on SWIG usage


aprokop:
The CI fails due to a brand new (as of today) move to a new CI server. Hang on.

PR #25, opened 2017-02-28
Fix link to documentation

youngmit:


aprokop:
ok to test

aprokop:
Thanks @youngmit. I should really be careful.

PR #28, opened 2017-02-28
ForTrilinos: added codecov badge

aprokop:


aprokop:
Alright, lets see if CI on develop works.

PR #47, opened 2017-05-02
Full SWIG automation and exception support

sethrj:
This updates the ForTrilinos wrappers to use SWIG in a fully automated fashion (rather than by hand). The 'ierr' integer variable in forteuchos is now automatically used for exception handling.

Closes #47


aprokop:
Now that I think about it, we should probably require that `-D ForTrilinos_ENABLE_DeveloperMode=ON` should require `MPI=ON`.

sethrj:
@aprokop I think this is ready to merge.

PR #49, opened 2017-05-09
Swig update

sethrj:
Fixes the .i files for the latest SWIG fortran (after changing things to get more SWIG unit tests to pass).

### DIFF
--- a/src/epetra/src/forepetraFORTRAN_wrap.cxx
+++ b/src/epetra/src/forepetraFORTRAN_wrap.cxx
@@ -173,6 +173,11 @@ template <typename T> T SwigValueInit() {
 
 
 
+/* Contract support */
+#define SWIG_contract_assert(nullreturn, expr, msg) if (!(expr)) { \
+swig::fortran_store_exception(SWIG_ValueError, msg); return nullreturn; }
+
+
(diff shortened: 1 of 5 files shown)


aprokop:
Why revert pedantic warnings? Did it break something for you?

sethrj:
If you pull the latest SWIG fortran develop version, you will see that your current version breaks, and that adding the semicolons back is required. The pedantic warnings were an indication that I didn't have the exception calling code written correctly in the SWIG .i files. I've fixed that.

aprokop:
I must be missing something. For me, the compiler produced a warning when it encountered `};` in the code. I am using gcc-6.3.1. I updated the swig, and regenerated the files, and it compiled fine for me.

Why do you need `do/while` for exceptions? It almost seems like you want to use the exception macro like `MACRO();` with semicolon after it, and that you add `do/while` only for that.

sethrj:
The SWIG convention is to write exception calls in the form:
```c++
        SWIG_exception(SWIG_RuntimeError, e.what());
```
So if you ever want to use the same `.i` files for e.g. Python as in Fortran, you will need to write the exception calls like that. The `do/while` is just to enforce the trailing semicolon; if you look through the SWIG library you'll see this is done by other languages too.

PR #50, opened 2017-05-09
Fix MPI_Comm treatment in simple interface

aprokop:
PR #47 was pushed even though I now discovered it did not treat MPI_Comm
right resulting in uknown types in .f90 files due to missing typemaps.
This patch fixes the issue.


sethrj:
Replaced by !53

PR #59, opened 2017-08-04
Fix sublists in ParameterList

sethrj:


aprokop:
Rebased on top of develop and regenerated the remaining wrapper files using the updated SWIG version.

Issue #10, opened 2017-02-03
Teuchos: wrap MPI communicator


aprokop:

aprokop:
In the simple_interface I simply used `integer` on Fortran side, which I then converted to `MPI_Comm` on C++ side through `Mpi_Comm_f2c`.

aprokop:
Related to #70.

PR #60, opened 2017-08-08
[WIP] Tpetra: nothing works, just saving the state

aprokop:


aprokop:
Need a proper treatment of `Teuchos::Comm<int>`. So, blocked by #10.

PR #62, opened 2017-08-17
Updating docker configuration

aprokop:


aprokop:
The build failure is about `x.mod` which is a different problem.

PR #93, opened 2017-11-08
Rename .f90 to F90

aprokop:
This is required for Doxygen as it is unable to parse comments in f90
files, only in F90.


aprokop:
This needs to wait until sethrj/swig#48 is fixed, otherwise in developer mode it would still generate `.f90`.

aprokop:
Closing as there is no reason for it right now.

Issue #80, opened 2017-10-16
Integrate Trilinos documentation into Fortran code


sethrj:
- [ ]  Export Trilinos Doxygen as XML
- [ ] Convert XML to SWIG .i file with [doxy2swig](https://github.com/m7thon/doxy2swig)
- [ ] Modify .i file as needed
- [x] (in SWIG fortran) add support for `%feature("autodoc")`

aprokop:
The way PyTrilinos currently works:
- In `$PYTRILINOS_DIR/CMakeLists.txt` it configures `PyTrilinos_DOCSTRINGS` (there are a few errors in it, though)
- In `$PYTRILINOS_DIR/docs/Doxygen` it a) generates helpers scripts from `.in` files using `CONFIGURE_FILE` command; b) Runs Doxygen for the specified packages to produce XML files; c) Runs `doxy2swig` file on those XML files to produce `.i` files with docstrings (which are stored in the binary directory/docs/Doxygen
- In `$PYTRILINOS_DIR/src` it a) includes that Doxygen binary directory; b) Add `-I` option to running `swig` so that it can pick up those files; and c) manually includes the docs files (`*_dox.i`) into the proper package files (like `Epetra.i`).

aprokop:
We need to copy this machinery from PyTrilinos, updated with a more recent version of `doxy2swig` (including @mattbement 's changes to include meta-data) and put it in ForTrilinos.

PR #101, opened 2017-11-16
Initial commit of doxy2swig

mattbement:
Following the pattern of PyTrilinos, creating a Doxygen directory which will eventually hold .i files for the wrapped packages.  This directory also holds the doxy2swig python script, which has been slightly modified to warn that .i files it produces are auto-generated.


aprokop:
Fix #80 when merged.

PR #102, opened 2017-11-16
Initial implementation of several Tpetra tutorials and supporting code/documentation

tjfulle:
@aprokop 

- Updated documentation based on my experience installing ForTrilinos
- Created Tpetra examples.  Located in src/tpetra/tutorials
- Added a fortpetra_types module.  This may be absorbed in to the fortpetra
  module.  It provides named constants for types used in Tpetra (rather than
  using the named types in the iso_c_binding module)
- Updates documentation for the Tpetra tutorials.

The tutorials also serve as tests.  If they build and run, they "pass" (there
are several correctness checks in each)

One tutorial, MapsAndVectors.f90 is still incomplete.  It is waiting on the
TpetraMap constructor that accepts a list of elements.


aprokop:
ok to test

docs/source/install.rst
```diff
@@ -92,6 +100,7 @@ required to build and install ForTrilinos:
 
           -D Trilinos_ENABLE_Amesos2=ON
           -D Trilinos_ENABLE_Anasazi=ON
+          -D Anasazi_ENABLE_Epetra=OFF
```
aprokop: This is not needed as it is done automatically by `Trilinos_ENABLE_Epetra=OFF` below.

docs/source/links.txt
```diff
@@ -0,0 +1,22 @@
+
+.. _Trilinos: http://www.trilinos.org/
+.. _Tpetra: http://www.trilinos.org/packages/tpetra
```
aprokop: I think there has been effort to move Trilnos to the new trilinos.github.io website. Should we link there instead?

src/tpetra/tutorials/MapsAndVectors.f90
```diff
@@ -0,0 +1,229 @@
...
+#else
+call comm%create()
+#endif
+
+my_rank = comm%getRank()
+num_procs = comm%getSize()
+
+! Tpetra objects are templated are templated on several parameters.  ForTrilinos
+! provides concrete specializations of Tpetra objects and exposes them to
+! Fortran users.  The module fortpetra_types provides named constants of type
```
aprokop: `fortpetra_types` is gone, fix the language.

src/tpetra/tutorials/MapsAndVectors.f90
```diff
@@ -0,0 +1,229 @@
...
+! default integer for Tpetra data types that can be used as the kind type
+! parameters.
+
+! Tpetra has local and global Maps.  Local maps describe objects that are
```
aprokop: The whole paragraph is very confusing to me. Tpetra has local and global indices, but not local and global maps. Can we instead use some language from Tpetra's Map doxygen documentation?

src/tpetra/tutorials/MapsAndVectors.f90
```diff
@@ -0,0 +1,229 @@
...
+! expects a size_type for this value.
+num_local_entries = 5
+
+! num_global_entries: The total (global, i.e., over all MPI processes) number of
+! entries (indices) in the Map.  ForTpetra expects global_size_t for this value.
+! This type is at least 64 bits long on 64-bit machines.
+!
+! For this example, we scale the global number of entries in the Map with the
+! number of MPI processes.  That way, you can run this example with any number
+! of MPI processes and every process will still have a positive number of
```
aprokop: I would say "and every process will still have at least one entry".

src/tpetra/tutorials/MapsAndVectors.f90
```diff
@@ -0,0 +1,229 @@
...
+
+! Tpetra can index the entries of a Map starting with 0 (C style),
+! 1 (Fortran style), or any base you want.  1-based indexing is
+! handy when interfacing with Fortran.  We choose 1-based indexing
+! here.
+index_base = 1
+
+! Create some Maps.  All Map constructors must be called as a collective over
+! the input communicator.  Not all Map constructors necessarily require
+! communication, but some do, so it's best to treat them all as collectives.
```
aprokop: Is that true? What are Map constructors that do not require global communication? I would assume that they all need `minGID` and `maxGID`?

src/tpetra/tutorials/MapsAndVectors.f90
```diff
@@ -0,0 +1,229 @@
...
+! Create a Map that puts the same number of equations on each processor.  The
+! resulting Map is "contiguous and uniform."
+call contig_map%create(num_global_entries, index_base, comm)
+
+! contig_map is contiguous by construction.  Test this at run time.
+EXPECT_TRUE(contig_map%isContiguous())
+
+! contig_map2: Create a Map which is the same as contig_map, but uses
+! a different Map constructor.  This one asks for the number of entries on each
+! MPI process.  The resulting Map is "contiguous" but not necessarily uniform,
```
aprokop: I don't think uniform means that exactly same number of entries per processor. I think they can differ by one.

src/tpetra/tutorials/MapsAndVectors.f90
```diff
@@ -0,0 +1,229 @@
...
+! contig_map, but distributes them differently, in round-robin (1-D cyclic)
+! fashion instead of contiguously.
+
+! We'll use the version of the Map constructor that takes, on each MPI process,
+! a list of the global entries in the Map belonging to that process.  You can
+! use this constructor to construct an overlapping (also called "not 1-to-1")
+! Map, in which one or more entries are owned by multiple processes.  We don't
+! do that here; we make a nonoverlapping (also called "1-to-1") Map.
+num_elements_per_proc = 5
+call element_list%create()
```
aprokop: This has to be changed for the latest changes in Map constructor that can take in the Fortran array.

src/tpetra/tutorials/MapsAndVectors.f90
```diff
@@ -0,0 +1,229 @@
...
+end do
+! TODO: TpetraMap%create does not yet take a list of elements
+! TODO: when the element list creation is complete, remove all lines starting
+! with "! MCELC (Map Creation from Element List Complete)
+!call cyclic_map%create(num_global_entries, element_list, index_base, comm);
+
+! If there's more than one MPI process in the communicator, then cyclic_map is
+! definitely NOT contiguous.
+! MCELC: if (num_procs > 1) then
+! MCELC:   x_test = .not. cyclic_map%isContiguous()
```
aprokop: Don't forget to uncomment all these.

src/tpetra/tutorials/PowerMethod.f90
```diff
@@ -0,0 +1,335 @@
...
+  write(*, *) "Increasing magnitude of A(1,1), solving again"
+
+! Must call resumeFill() before changing the matrix, even its values.
+call A%resumeFill()
+
+id_of_first_row = 1
+if (map%isNodeGlobalElement(id_of_first_row)) then
+  ! Get a copy of the row with with global index 1.  Modify the diagonal entry
+  ! of that row.  Submit the modified values to the matrix.
+  print*, 'here i am!'
```
aprokop: Don't forget to remove the debugging code.

src/tpetra/tutorials/PowerMethod.f90
```diff
@@ -0,0 +1,335 @@
...
+
+! Increase diagonal dominance
+if (my_rank == 0) &
+  write(*, *) "Increasing magnitude of A(1,1), solving again"
+
+! Must call resumeFill() before changing the matrix, even its values.
+call A%resumeFill()
+
+id_of_first_row = 1
+if (map%isNodeGlobalElement(id_of_first_row)) then
```
aprokop: Would be nice to explain why we need to modify the first row. Also, why do we not modify the last one too?

docs/source/install.rst
tjfulle: Ok.

docs/source/links.txt
tjfulle: Ya, that is a good idea.

src/tpetra/tutorials/MapsAndVectors.f90
tjfulle: doh!  Missed that one.

src/tpetra/tutorials/MapsAndVectors.f90
tjfulle: ok

src/tpetra/tutorials/PowerMethod.f90
tjfulle: doh!  good catch

src/tpetra/tutorials/MapsAndVectors.f90
tjfulle: A local map is a map that is replicated across all participating MPI processes and a global map is one that is distributed.  Perhaps there is a better description, but that is how it is described in the Tpetra documentation and tutorials.

docs/source/install.rst
tjfulle: I remember why I add this, I was trying to get past the compile error that was fixed when you reminded me about the patches.

src/tpetra/tutorials/MapsAndVectors.f90
tjfulle: Maps that take a list of indices may not require communication.  Global quantities can be computed as needed.  The Tpetra docs say something like "Tpetra reserves the right to use MPI collectives to check input values in a debug build"

src/tpetra/tutorials/MapsAndVectors.f90
tjfulle: I believe that is true, but can differ by one on only one processor.  The docs say something like uniformity is only guaranteed if the map is constructed with a constructor that enforces uniformity.

src/tpetra/tutorials/MapsAndVectors.f90
tjfulle: done

src/tpetra/tutorials/MapsAndVectors.f90
tjfulle: done

src/tpetra/tutorials/PowerMethod.f90
tjfulle: The matrix doesn't *need* to be modified.  Modifying the matrix makes it more diagonally dominant and the power method converges faster.  The point of modifying was just to demonstrate modifying a matrix, but I can put in a comment.

tjfulle: As to why not modify the last - because I just copied the Tpetra tutorial and in that tutorial it only modified the first.

src/tpetra/tutorials/MapsAndVectors.f90
aprokop: I don't know if there is a good definition of uniformity. And no, it can be off by 1 on multiple processors, for instance, try constructing a map with 6 global element on 4 processors. Two of them will get 1, and two will get 2.

src/tpetra/tutorials/PowerMethod.f90
aprokop: OK, that's fine.

src/tpetra/tutorials/MapsAndVectors.f90
tjfulle: Oh, right.

tjfulle: Most of the commentary is copied straight from the Tpetra tutorial.  I'd rather keep it as is, or modify both this commentary and the Tpetra commentary so they are consistent *or* strip the commentary away from fortpetra and leave just the stuff that describes fortpetra.  There is already a note in the html that the tutorials are not meant to replace their Tpetra counterparts but only show the implementation in fortpetra.

docs/source/packages/tpetra/overview.rst
```diff
@@ -0,0 +1,57 @@
...
+  differentiation objects from the Sacado package, or stochastic PDE
+  discretization types from the Stokhos package.
+* Support for many different shared-memory parallel programming models
+
+Tpetra Support in ForTrilinos
+-----------------------------
+
+Tpetra is a large library and its full functionality has not been exposed in
+ForTrilinos, though coverage is actively being expanded.  The Tpetra
+tpetra_examples_ provide a good overview of the Tpetra functionality currently
```
aprokop: This does not compile:
```
/home/xap/code/trilinos/fortrilinos/packages/ForTrilinos/docs/source/packages/tpetra/overview.rst:52: ERROR: Unknown target name: "tpetra_examples".                                           
```

tjfulle: hmmm, `tpetra_examples_` is defined at the top of `/docs/source/packages/tpetra/examples/index.rst`.  Can you think of why it is not picked up?

aprokop: According to [this](https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#internal-and-external-links), `link_` only works within the same file. For external files one should use ``:ref:`link` ``.

PR #106, opened 2017-11-20
Adding comprehensive tests for TpetraMap

tjfulle:
@aprokop, I added a comprehensive suite of tests for `TpetraMap`.  I believe (nearly) every public function in `fortpetra` is tested.  There are 2-3 functions that are not tested, they are labeled `TODO`.  Some probably don't need to be included.  If you could go through and read the `TODO`'s and provide feedback, that would be great.


src/tpetra/test/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@
 
 SET(TESTS
   test_tpetra_map.f90
+  test_TpetraMap.f90
```
aprokop: This can probably replace whatever is in `test_tpetra_map.f90`.

tjfulle:
Ya, it exercises every use case in the first text file

tjfulle:
I think I know the source of the test error, I’ll get on it first thing in the morning

tjfulle:
@aprokop, fix for serial is in.

aprokop:
@tjfulle You used `comm%getSize() == 0` in a bunch of places instead of `comm%getRank() == 0` :)

tjfulle:
@aprokop, good catch!  Just pushed the fix.

tjfulle:
I also just noticed some bad logic with the error flag.  It was globally set based on the pass/fail status of the *last* test.  I added a new global error flag to avoid that situation

aprokop:
This looks good! The only thing the remains on my mind is whether we should also put `EXPECT_EQ(ierr, 0)` after each Tpetra call. But I think that's unnecessary, as long as we have other tests that test exceptions. I think the only situation where it would be necessary is when we need `TEUCHOS_TEST_THROW`. Do we have situations like that?

tjfulle:
I'm not sure ?  I would probably prefer that each subtest be a standalone test that throws on failure.  But, if I am not mistaken, that would require every subtest to be it's own `program` and would require far more files to maintain.  I opted to go with a global error counter and only throw at the end so that each individual subtest had a chance to run.

tjfulle:
The problem with a `EXPECT_EQ` invocation after each Tpetra call is that if one test failed, then any test thereafter would not run.  Each test is independent, so we would want to avoid that.

aprokop:
Lets consider, for instance, `test_isDistributed`. If you are able to construct `Obj` but `isDistributed` throws (in C++), it will set `ierr` to 1 on the Fortran side. However, this will be unnoticed by `ierr2`, and the default value for `fresult` will be used which may pass the test. So, I think the test counter `ierr2` should be updated as `ierr2 = ierr2 + ierr` after each call. What do you think?

tjfulle:
Yes!  I totally spaced that there was a global `ierr` variable.  A test skeleton should be something like:

```
integer function test_TestName()
  ...
  call Obj%create(...args...)
  if (ierr /= 0) then
    ierr2 = err
    return
  end if

  ! Do the test, checking for ierr where appropriate

  call Obj%release()

  test_<TestName> = ierr2 + err
  return
end function test_TestName
```

Does that cover the cases you were thinking?

aprokop:
Yes! Though, I think `ierr2 = ierr; return` should rather be `test_<testName> = ierr; return`. And the only situation where the pattern is different would be the test where we expect the thing to throw (I think `CrsMatrix` tests would test that, for instance trying to insert into a non-owned row).

And sorry, it really is our fault about the global `ierr`, it's not documented anywhere except our heads.

tjfulle:
Ok, I'm finishing up changing the map tests and I incorporated the changes in the test generation script for the other classes.

tjfulle:
@aprokop, on further thought, `ierr` should be reset to 0 after each subtest (so that future subtests do not wrongly fail due to `ierr /= 0` from a previous test).  the skeleton I have implemented is roughly:

```
integer function test_TestName()
  integer :: jerr
  ...
  call Obj%create(...args...)
  if (ierr /= 0) then
    test_TestName = jerr + err
    ierr = 0
    return
  end if

  ! Do the test, checking for ierr where appropriate

  call Obj%release()

  test_TestName = jerr + ierr
  ierr = 0
  return
end function test_TestName
```
```

**Comment by aprokop** [comment]
You could simplify it further by setting `ierr` to 0 outside the test.

**Comment by tjfulle** [comment]
I would prefer the test that triggered `ierr` to be responsible for reseting to to 0.  Any time `ierr` is not 0 *outside* of a subtest, it would mean something globally went wrong, I think?

**Comment by aprokop** [comment]
OK. Let me know when it's ready for another glance.

**Comment by tjfulle** [comment]
@aprokop, the latest push has the above changes made.

**Comment by tjfulle** [comment]
@aprokop, I don't particularly like the current way of doing each subtest and would prefer a macro!

I thought about checking `ierr` after every Tpetra call, but opted for just the constructors in this case.  A macro could wrap each call to Tpetra that would make the call, check the error state, and return if something went wrong:

```
#define CALL_AND_CHECK_IERR(NAME, PROCEDURE, ...) \
  call PROCEDURE(##__VA_ARGS__) \
  if (ierr /= 0) then \
    NAME = ierr \
    return \
  end if

CALL_AND_CHECK_IERR(test_TestName, Obj%procedure, arg1, arg2, ..., argn)
```

would expand to

```
call Obj%procedure(arg1, arg2, ..., argn)
if (ierr /= 0) then
  test_testName = ierr
  ierr = 0
  return
end if
```

**Warning** my C macro programming foo is very low

aprokop:
I think this would be a fine approach. My only concern is whether macros are portable in Fortran. Do you know if they are?

tjfulle:
I am actually just finishing a quick initial set of macros, I'll push so you can see.  Variadic macros are not supported, so I had to create several macros to cover different use cases.

tjfulle:
@aprokop Okay, I just pushed a commit with some testing macros.  I only used the macros in one test.  If you think it is cleaner, I can adopt the macros in my test generation script and convert the rest of the tests in `test_tpetra_map.f90` to use the macros.

aprokop:
@tjfulle I like `ADD_TEST`. I am less excited about other macros. What if we move `PROCEDURE` out of the macro, and just have `if (ierr /= 0)` block in it? I think this would make it clear.

One more thing I completely forgot about: would tests actually work in parallel? What happens if some condition passes on rank 0, but fails on other ranks? I think Teuchos test macros aggregate results, but here we don't seem to?

tjfulle:
@aprokop, ya, those `CALL_AND_...` macros were ugly.  The updated version implements two new macros: `TEST_FOR_IERR(TEST_NAME)` and `SET_ERROR_COUNT_AND_RETURN`.  I think the code cleans up quite a bit with these.

As for synchronizing error counts in parallel, that is not being done.  How would you suggest?  I looked at the `CHECK_EQ` macro and didn't notice anything, but it does call an `Insist` function that might implement some kind of parallel reduction?

aprokop:
The macros look much nicer now, thank you for updating those.

I think we could updated `ADD_TEST` to do a `gatherAll` or a reduction (sum?) of the returned errors, and check that on the rank 0. This should be sufficient. The only downside is that the error messages *inside* a test may hide the error if it happens on a different rank.

tjfulle:
Ya, it's really too bad that fortran macro expansion doesn't support strings stringification, that would help collate errors, rather than just writing everything to the console.

tjfulle:
@aprokop , the map tests have been updated to use the macros.  An issue with the doing a reduction in `ADD_TEST` is that fortran doesn't allow locally scoped variables.  So, the test would have to declare, say, `global_err` and `local_err`.  Then `ADD_TEST` could do `local_err = TEST_NAME(); MPI_reduceAll(...)` to get the `global_err`.  `local_err` and `global_err` could be passed to `ADD_TEST`, or `ADD_TEST` would just use them and a test wouldn't compile unless those variables existed.  Same goes for `comm`.

Alternatively, we could write a new header that defines the variables required by `ADD_TEST`.  It would have to be in a new header and that header would have to be `#include`d *after* all of the `use` statements.

tjfulle:
@aprokop latest push has some pretty significant changes to the testing macros.  Feedback is appreciated.  Not having locally scoped variables in Fortran makes things a little messier than I'd like, but I think the final solution works well (see the maps tests)

aprokop:
I'm not sure, why did you put all the macros in `hpp.in`? Isn't it sufficient to just have `#cmakedefine HAVE_MPI` there, and then include the generated header file into existing `FortranTestMacros.h`?

Otherwise, looks pretty good. I think `__FUNCTION__` is not supported in Fortran, so we'll need to explicitly type `TEST_FOR_IERR(<name>)`.

I would also move `implicit none` out of `DECLARE_TEST_VARIABLES`.

tjfulle:
There is a very simple reason for the way the files were grouped - it was my first foray in to CMake :)

I don't have a strong preference on the `implicit none`.  I put it in the macro with the intent that the test declaration macro would *always* be the first declaration and it assured that `implicit none` is always set.

aprokop:
OK, just leave `implicit none` in the macro. But please separate `hpp.in` and the `.h` as it was before.

tjfulle:
@aprokop said:
> But please separate hpp.in and the .h as it was before

Done.

aprokop:
Where?

tjfulle:
On my machine :)  I haven't pushed yet

PR #112, opened 2017-11-27
WIP: TpetraMultiVector, TpetraCrsMatrix tests; Updated testing macros

tjfulle:
@aprokop, get in touch with me when you are back about this PR.  I imagine it'll take an 30 or so minutes from us to walk through it.  There are many `TODO`s that need to get done on the swig side of things.

### DIFF
--- a/src/utils/src/FortranTestMacros.h
+++ b/src/utils/src/FortranTestMacros.h
@@ -154,7 +154,18 @@ use DBCF_M
   if (COMM_RANK == 0) then;                                       \
   WRITE(0, '(A,A,A,I6)') "File '", __FILE__, "', line ", __LINE__; \
   write(0, '(A)') "  TEST_COMPARE_FLOATING_ARRAYS(ARR1,ARR2,TOL)";\
-  write(0, '(A)') "Error: Expected ARR1 == ARR2";                 \
+  write(0, *) "Error: Expected ", ARR1, "==",  ARR2;              \
+  end if;                                                         \
+  success = .false.; ierr = 0;                                    \
+  return;                                                         \
+  end if
+
+#define TEST_COMPARE_ARRAYS(ARR1, ARR2)                           \
+  IF(ABS(MAXVAL(ARR1 - ARR2)) > epsilon(1.)) THEN;           \
+  if (COMM_RANK == 0) then;                                       \
+  WRITE(0, '(A,A,A,I6)') "File '", __FILE__, "', line ", __LINE__; \
+  write(0, '(A)') "  TEST_COMPARE_ARRAYS(ARR1,ARR2)";             \
+  write(0, *) "Error: Expected ", ARR1, "==",  ARR2;              \
   end if;                                                         \
   success = .false.; ierr = 0;                                    \
   return;                                                         \
@@ -165,7 +176,7 @@ use DBCF_M
   if (COMM_RANK == 0) then;                                       \
   WRITE(0, '(A,A,A,I6)') "File '", __FILE__, "', line ", __LINE__; \
   write(0, '(A)') "  TEST_ARRAY_EQUALITY(ARR1,ARR2,TOL)";         \
-  write(0, '(A)') "Error: Expected ARR1 == ARR2";                 \
+  write(0, *) "Error: Expected ", ARR, "==",  VAL;                \
   end if;                                                         \
   success = .false.; ierr = 0;                                    \
   return;                                                         \
@@ -175,8 +186,8 @@ use DBCF_M
   if(abs(maxval(ARR - VAL)) <= TOL) then;                         \
   if (COMM_RANK == 0) then;                                       \
   WRITE(0, '(A,A,A,I6)') "File '", __FILE__, "', line ", __LINE__; \
-  write(0, '(A)') "  TEST_ARRAY_INEQUALITY(ARR1,ARR2,TOL)";       \
-  write(0, '(A)') "Error: Expected ARR1 /= ARR2";                 \
+  write(0, '(A)') "  TEST_ARRAY_INEQUALITY(ARR,VAL,TOL)";         \
+  write(0, *) "Error: Expected ", ARR, "/=",  VAL;                \
   end if;                                                         \
   success = .false.; ierr = 0;                                    \
   return;                                                         \
--- a/src/utils/src/FortranTestMacros.h
+++ b/src/utils/src/FortranTestMacros.h
@@ -1,3 +1,4 @@
+! vi: ft=fortran
 #ifndef FOTRANTESTMACROS_H
 #define FOTRANTESTMACROS_H
 
@@ -5,6 +6,7 @@
 /* Options */
 
 #include "DBCF.h"
+use DBCF_M
 
 #define ASSERT_C_ASSOC( X ) \
     Insist( C_ASSOCIATED( X % instance_ptr ), "X not C_ASSOCIATED" )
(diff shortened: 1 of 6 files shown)


src/utils/src/FortranTestMacros.h
```diff
@@ -41,81 +43,216 @@
...
-#define DECLARE_TEST_VARIABLES() \
-  use mpi; \
-  implicit none; \
-  integer ERR1(1), ERR2(1), IERROR, COMM_RANK, ERROR_COUNTER
+#define DECLARE_TEST_VARIABLES()                                  \
+  use mpi;                                                        \
+  implicit none;                                                  \
+  logical LOCAL_SUCCESS;                                          \
+  character(len=256) JNKSTR;                                      \
+  integer ERR1(1), ERR2(1), IERROR;                               \
```
aprokop: Preferably more clear names for `ERR1` and `ERR2`? Like `ERR_LOCAL` and `ERR_GLOBAL`?

src/utils/src/FortranTestMacros.h
```diff
@@ -41,81 +43,216 @@
...
 #else
 
-#define DECLARE_TEST_VARIABLES() \
-  implicit none; \
-  integer ERROR_COUNTER
+#define DECLARE_TEST_VARIABLES()                                  \
+  implicit none;                                                  \
+  logical LOCAL_SUCCESS;                                          \
+  character(len=256) JNKSTR;                                      \
+  integer COMM_RANK, TOTAL_TESTS, FAILED_TESTS
```
aprokop: Why do we use `COMM_RANK` in a `#else` branch where there is no MPI?

src/utils/src/FortranTestMacros.h
```diff
@@ -41,81 +43,216 @@
...
+  write(0,'(A,I3,A)') "100% of ", TOTAL_TESTS, " tests PASSED";   \
+  else;                                                           \
+  write(*,'(I3,A,I3,A)') FAILED_TESTS, " of ", TOTAL_TESTS, " tests FAILED";   \
+  Insist(.false., "FAILED TESTS ENCOUNTERED" );                   \
   end if
 
 #endif
 
+! The TEST_* macros to follow are intended to be called from within
+! a FORTRILINOS_UNIT_TEST.  Insetad of stopping calculations at the first sign of
```
aprokop: Insetad -> Instead

src/utils/src/FortranTestMacros.h
aprokop: OK, I see that you use it in `TEST_IERR` (which is MPI agnostic), but it still is not needed in `ADD_SUBTEST_AND_RUN`.

src/utils/src/FortranTestMacros.h
```diff
@@ -41,81 +43,216 @@
...
+  if (COMM_RANK == 0) then;                                       \
+  WRITE(0, '(A,A,A,I6)') "File '", __FILE__, "', line ", __LINE__; \
+  write(0, '(A)') "  TEST_IERR()";                                \
+  write(0, '(A)') "Error:", trim(serr);                           \
+  end if;                                                         \
+  success = .false.; ierr = 0;                                    \
+  return;                                                         \
+  end if
+
+#define TEST_COMPARE_FLOATING_ARRAYS(ARR1, ARR2, TOL)             \
```
aprokop: This seems to be a duplicate of `TEST_ARRAY_EQUALITY`.

src/utils/src/FortranTestMacros.h
```diff
@@ -41,81 +43,216 @@
...
+  WRITE(0, '(A)') "  TEST_NOTHROW(CODE)";                         \
+  WRITE(0, '(A)') "Error: Expected CODE to not set ierr /= 0";    \
+  WRITE(0, '(A,A)') "serr: ", trim(serr);                         \
+  end if;                                                         \
+  success = .false.; ierr = 0;                                    \
+  return;                                                         \
+  end if
+
+#define OUT0(STRING)                                              \
+    if (COMM_RANK == 0) WRITE(0, *) STRING
```
aprokop: Maybe use this inside other macros?

src/utils/src/FortranTestMacros.h
```diff
@@ -41,81 +43,216 @@
...
-
-#define TEST_FOR_IERR( NAME ) \
-    if (ierr /= 0) then; \
-      NAME = ierr; \
-      ierr = 0; \
-      return; \
-    endif
+    if(.NOT. ierr == 0 ) then;                                    \
+    WRITE( 0, * ) "Expected ierr = 0, but got ", ierr;            \
+    WRITE( *, * ) "With associated serr = ", trim(serr);          \
```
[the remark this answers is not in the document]
tjfulle: It should be `WRITE(0, *)`.  I'll get that fix in

src/utils/src/FortranTestMacros.h
```diff
@@ -41,81 +43,216 @@
...
+    Insist(.false., "Expected ierr = 0" );                        \
+    end if
```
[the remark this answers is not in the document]
tjfulle: I think `CHECK_IERR` should be removed from `FortranTestMacros.h` and be put in a "non-test" include.  It could be something like

```
WRITE(JNKSTR,'(A,I3,A)')"Expected ierr=0, but got ",ierr,NEW_LINE('A')//"Associated serr: "//TRIM(serr);Insist(.FALSE., JNKSTR)
```

This macro is 127 characters and could be used in all of ForTrilinos.

The other macros may be able to be made less than 132 characters.  The hardest would be the `ADD_AND_RUN_SUBTEST` macro, it would have to be modified to make the value of `success` not be collected over all processes, meaning each test would be in charge of doing the reduction.  I don't think this is such a bad idea.  `Teuchos` doesn't do reductions on `success`, it assumes the test does it (I believe?).

src/utils/src/FortranTestMacros.h
```diff
@@ -41,81 +43,216 @@
...
-    ierr = 0; \
-    return
+#define FORTRILINOS_UNIT_TEST(NAME)                               \
+    subroutine NAME(success);                                     \
+    implicit none;                                                \
+    logical :: success
 
-use DBCF_M
+#define END_FORTRILINOS_UNIT_TEST(NAME)                           \
+    if (success) success = (ierr == 0);                           \
```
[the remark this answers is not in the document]
tjfulle: ok

src/utils/src/FortranTestMacros.h
tjfulle: ok

src/utils/src/FortranTestMacros.h
```diff
@@ -41,81 +43,216 @@
...
-    ERR1(1) = TEST_NAME(); \
+#define ADD_SUBTEST_AND_RUN(NAME)                                 \
+    LOCAL_SUCCESS = .true.;                                       \
+    TOTAL_TESTS = TOTAL_TESTS + 1;                                \
+    call NAME(LOCAL_SUCCESS);                                     \
+    if (LOCAL_SUCCESS) then;                                      \
+    ERR1(1) = 0;                                                  \
+    else;                                                         \
+    ERR1(1) = 1;                                                  \
+    end if;                                                       \
```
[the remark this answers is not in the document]
tjfulle: ok

src/utils/src/FortranTestMacros.h
```diff
@@ -41,81 +43,216 @@
...
     call MPI_ALLREDUCE(ERR1, ERR2, 1, MPI_INTEGER, MPI_MAX, MPI_COMM_WORLD, IERROR); \
-    if ( ERR2(1) /= 0 ) then; \
-      if (COMM_RANK == 0) write(0, *) "Test FAILED!"; \
-      ERROR_COUNTER = ERROR_COUNTER + 1; \
-    endif
+    if ( ERR2(1) /= 0 ) then;                                     \
+    if (COMM_RANK == 0) write(0, *) "Test FAILED!";               \
```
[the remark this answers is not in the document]
tjfulle: Yes!

src/utils/src/FortranTestMacros.h
tjfulle: ok

src/utils/src/FortranTestMacros.h
tjfulle: We could, but then getting line lengths under 132 characters would be really hard

src/utils/src/FortranTestMacros.h
tjfulle: yes, it is.  `Teuchos` provides both, but in Fortran both aren't necessary.  I included both to more easily copy/paste C++ tests and port them to Fortran

aprokop:
@tjfulle Can you also please separate this PR into three different ones: one with Fortran macros update + Map changes, one with MultiVector, and one with CrsMatrix? This way, I can push my related changes to corresponding branches and merge independently.

src/utils/src/FortranTestMacros.h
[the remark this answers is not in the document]
tjfulle: Actually, this won't work in fortran since implicit conversion from `LOGICAL` to `INTEGER` is not allowed.  I'll use `ERR(1) = MERGE(1, 0, LOCAL_SUCCESS)` instead.

src/utils/src/FortranTestMacros.h
tjfulle: Yep, this is one of the areas where non-locally scoped variables create a need for ugly workarounds.

src/utils/src/FortranTestMacros.h
tjfulle: also, `TEST_ARRAY_EQUALITY` does not take an explicit tolerance.

aprokop: ? Line 174: `#define TEST_ARRAY_EQUALITY(ARR, VAL, TOL)`

src/utils/src/FortranTestMacros.h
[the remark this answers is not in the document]
tjfulle: This is what I can find on allowed line lengths from different compiler vendors:

Vendor | Allowed Line Length | Modifiable?
------- | -------------------- | -----------
nag | 132 | no
pgi | 264 | no
gnu | 132 | yes (`-ffree-line-length-n`)
intel | 2048 | no
flang | 132 | no
ibm | 132 | no
sun | 132 | no

src/utils/src/FortranTestMacros.h
tjfulle: Doh! You are right.  I probably only added the two different macros to mimic the functionality of `Teuchos`.

src/utils/src/FortranTestMacros.h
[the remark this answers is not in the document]
tjfulle: I'm making some changes that will get all macros to emit code < 132 characters in length (per line)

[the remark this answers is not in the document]
aprokop: Unfortunately, this also means no `__FILE__` in macros, which would make test failures more cryptic.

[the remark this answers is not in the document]
tjfulle: Do the macros in the `DBCF.h` compile on different compilers?  They all use `__FILE__` and `__LINE__`.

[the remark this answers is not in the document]
aprokop: I think I was able to compile ForTrilinos using Flang once I replaced `__FILE__` by `""`.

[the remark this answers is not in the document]
tjfulle: Wait, it looks like they have different definitions for different compilers.  Yuck!

tjfulle:
@aprokop, I've got a good solution that gets by the line length limitations and, in fact, is more general purpose.  I'll test and push tomorrow morning.

tjfulle:
This PR will be split up in to separate PRs for the Macros/Maps, CrsMatrix, and MultiVector portions

PR #116, opened 2017-11-29
Tpetra: remove indexBase from arguments (hardcode to 1)

### LINKED ISSUE #103: ForTpetra: Index base argument to Map could cause headaches
@aprokop and I just discussed the following issue offline and thought it good to document it here.  Perhaps @mhoemmen also has an opinion?

`Tpetra::Map` objects take `indexBase` as an argument.  `indexBase` is required to be the minimum global id (GID) and defaults to 0 in C++ codes.  In ForTpetra, it is typical to use `indexBase` equal to 1, since Fortran array indexing starts at 1 - but that is only the default behavior.  Fortran arrays can be defined over an arbitrary range:

```fortran
real(kind=the_kind) :: a(10), b(-5:5)
```
both `a` and `b` have the same dimension, but different index ranges.  I'm not sure how the array `b` would look compared to `a` when passed to C++.

I believe ForTpetra has hardcoded some conversions to go back and forth between 0 C based indexing and 1 Fortran based indexing.  But, if a user defined the Fortran array with arbitrary range, the hard coded conversions could be incorrect.

I suggest removing the `indexBase` option from ForTpetra all together and requiring it always be the Fortran default of 1.  1 could then be hard-coded in the C++ wrappers and sent to Tpetra.  Further, it should be documented that ForTrilinos assumes that *all* Fortran arrays are indexed from 1.

aprokop:
Fix #103.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -34,12 +34,12 @@ Now is the extensive list.
 
 Status | Command| Comment
 -------|--------|---------
-:white_check_mark: | `Map(numGlobalElements, indexBase, comm)`
-:white_check_mark: :star: | `Map (global_size_t numGlobalElements, GlobalOrdinal indexBase, const Teuchos::RCP< const Teuchos::Comm< int > > &comm, LocalGlobal lg=GloballyDistributed, const Teuchos::RCP< Node > &node=defaultArgNode< Node >())` | No `Node` version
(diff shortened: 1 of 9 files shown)


tjfulle:
Can we hold off on this until after a PR for the test macros and map is merged?  It’ll save a huge headache!

aprokop:
@tjfulle OK.

PR #118, opened 2017-11-29
Updated testing framework and TpetraMap tests

tjfulle:
@aprokop 

Test Macros
-----------
- New test macros to make tests cleaner and easier to write.
- Error diagnostics from test macros are much better
- Macros interact with the fortest.mod modules
- All macros are < 132 characters long

TpetraMap Tests
---------------
- Updated to work with new test macros

This PR supersedes #112, which I will close shortly


aprokop:
Will review now. For the moment, please add copyright notices back.

aprokop:
@tjfulle Yes. You haven't notices because I added them just yesterday :)

aprokop:
Per @tjfulle 

> There is still an issue with `ierr` and `serr` and that is that they are only pulled in through `forteuchos`.  I wanted to make them independent of `forteuchos` so that the testing macros could get/set them without `forteuchos`, but that would require some swig work.  So, for now everything works, but the `TEST_*` macros in `FortranTestingMacros.h` have an implicit assumption of `forteuchos` being `use`d.  I don't `use forteuchos` in `FortranTestingMacros.h`, so it is implied that is `used` in the test file that includes `FortranTestingMacros.h`.  If `ierr` and `serr` were not in `forteuchos` but some other `for*` module, I would include that module in `FortranTestingMacros.h` and then the code paths would be more explicit.

tjfulle:
Copyright assertions added

tjfulle:
@aprokop , I'll open a new issue concerning `ierr` and `serr`

src/utils/src/fortest.f90
```diff
@@ -0,0 +1,1059 @@
...
+    ! ------------------------------------------------------------------------ !
+
+    ! Initialize module level variables
+    local_error(1) = 0
+    global_error(1) = 0
+    failed_tests = 0
+    total_tests = 0
+
+    ! Initialize MPI
+    ierror = 0
```
aprokop: `ierror = 0` has to go inside `HAVE_MPI`.

tjfulle:
The new test macros are merely wrappers to procedures in `fortest.f90`.  The reduction in line width in each test macro (to stay under 132 characters) led to some bloat in `fortest.f90`.  Specifically, procedures for comparing different arguments had to be written for each expected `kind` of argument.

src/utils/src/fortest.f90
```diff
@@ -0,0 +1,1059 @@
...
+                   fortest_array_equality_long1, &
+                   fortest_array_equality_long2
+end interface
+
+interface fortest_array_inequality
+  module procedure fortest_array_inequality_double1, &
+                   fortest_array_inequality_double2, &
+                   fortest_array_inequality_int1, &
+                   fortest_array_inequality_int2, &
+#ifdef SIZE_T_NOT_LONG
```
aprokop: Can you please add a few words on the macros `SIZE_T_NOT_LONG` and friends? I'm not sure where they are coming from, and what's their role.

src/utils/src/FortranTestMacros.h
```diff
@@ -2,124 +2,137 @@
...
-#define ASSERT_C_ASSOC2( X, Y )                                 \
-    Insist( C_ASSOCIATED( X % instance_ptr, Y % instance_ptr ), \
-            "X not C_ASSOCIATED with Y" )
-#define ASSERT_NOT_C_ASSOC2( X, Y )                                 \
-    Insist(.NOT.C_ASSOCIATED( X % instance_ptr, Y % instance_ptr ), \
-           "X C_ASSOCIATED with Y" )
+use DBCF_M
+use fortest
+
+#define FILENAME __FILE__
```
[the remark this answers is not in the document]
tjfulle: You are right, it will expand to the wrong filename.  My intent was that for compilers that did not support >132 characters, it would be only one place to change `FILENAME`.

src/utils/src/FortranTestMacros.h
```diff
@@ -2,124 +2,137 @@
...
+ CHECK_IERR(); \
+ CALL TEARDOWN_TEST2()
+
+#define FORTRILINOS_UNIT_TEST(NAME) \
+ SUBROUTINE NAME(SUCCESS); \
+ IMPLICIT NONE; \
+ LOGICAL :: SUCCESS
+
+#define END_FORTRILINOS_UNIT_TEST(NAME) \
+ SUCCESS = SUCCESS .AND. (IERR == 0); \
```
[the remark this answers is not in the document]
tjfulle: this is an assignment from `logical` to `logical`.  it is the case of `integer` to `logical` that did not work

src/utils/src/fortest.f90
tjfulle: Yes, I'll add some comments.  Essentially, with `gfortran` `c_size_t`, `c_long`, and `c_long_long` are all the same `kind`, so the compiler cannot overload a function whose only difference is the arguments are one of those integer kinds.

tjfulle:
@aprokop, okay, I addressed your comments and all tests build/pass now.  If you have more comments, let me know.  When this PR is merged I'll open (separate) PRs for the MultiVector and CrsMatrix tests

Issue #121, opened 2017-11-29
TpetraMultiVector test coverage


tjfulle:

tjfulle:
PR #122 goes a long way to getting pretty complete test coverage for `TpetraMultiVector`.  There are many "skeleton" tests in `test_tpetra_multivector.f90` that are commented out that can be activated by uncommenting the skeleton and completing the implementation.

aprokop:
We can start uncommenting tests.

aprokop:
Fixed by #262.

PR #122, opened 2017-11-29
Tpetra: MultiVector tests

tjfulle:
@aprokop 

All tests now pass

One TODO: `Get1dCopy` could take fortran array instead of `ArrayView`

Related issue: #121

### DIFF
--- a/src/tpetra/test/CMakeLists.txt
+++ b/src/tpetra/test/CMakeLists.txt
@@ -9,6 +9,7 @@
 
 SET(TESTS
   test_tpetra_map.f90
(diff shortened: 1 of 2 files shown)


tjfulle:
`test_tpetra_multivector.f90` has several test "skeletons" that are commented out that need to be uncommented and made to run.  However, most are tests for deprecated and/or "expert" methods in Tpetra.  Nevertheless, I think it is a good idea to merge the PR so that `TpetraMultiVector` test coverage will increase.  Future pushes can uncomment and finish the few remaining tests.

PR #123, opened 2017-11-29
Documentation of test macros

tjfulle:


src/utils/src/FortranTestMacros.h
```diff
@@ -9,6 +9,63 @@
...
+!
+!   The SETUP_TEST macro initializes module variables in the fortest module and
+!   initializes MPI for MPI builds of ForTrilinos (meaning that MPI should not
+!   be initialized elsewhere in the test!)
+!
+! - Adding and running subtests.  Use the macro ADD_SUBTEST_AND_RUN:
+!
+!      ADD_SUBTEST_AND_RUN(TestName)
+!
+!   where TestName is the name of a test defined in the program unit.  The test
```
aprokop: Do we require that the test name be a valid Fortran variable name? If yes, put some comment about this.

src/utils/src/FortranTestMacros.h
```diff
@@ -9,6 +9,63 @@
...
+!   should be defined using the FORTRILINOS_UNIT_TEST and
+!   END_FORTRILINOS_UNIT_TEST macros (after the program's "contains" directive):
+!
+!      FORTRILINOS_UNIT_TEST(TestName)
+!        ! Test definition
+!      END_FORTRILINOS_UNIT_TEST(TestName)
+!
+!   These macros define a subroutine named "TestName" and declare an in/out
+!   logical "SUCCESS" that can be set to .FALSE. if the test fails.
```
aprokop: Put some wording about local vs global success when run with MPI.

tjfulle:
@aprokop, I guess it’s implied in the statement that the macro creates a subroutine as the same name as the test.  I can add something more explicit.

aprokop:
Did not notice that sentence later.

PR #124, opened 2017-11-30
Progress in CrsMatrix

aprokop:


tjfulle:
@aprokop, did you completely do away with the interface that takes `nument`, `cols`, `vals` to the different `*[Local,Global]Values` functions?  I was just thinking that that interface might be useful for the case that a user is using an assumed size array (`real(kind) :: a(*)`) as opposed to an assumed shape array (`real(kind) :: a(:)`).  For the former case, the intrinsic `size` does not work and the size must be passed as an argument.  But, it seems that most of the functions in `tpetraFORTRAN_wrap.cxx` work strictly with assumed shape arrays, so this may be a non-starter.

aprokop:
@tjfulle For now, yes, I got rid of all `nument` functions. I do not know well the difference between two. Does it mean that all functions in ForTrilinos where we take in an assumed-shape array won't work with assumed-size arrays? Do we need to duplicate those functions to support that?

The above warrants the answer. But unless the solution is trivial, I don't think it's feasible to implement both in this release cycle.

tjfulle:
It was just a thought.  I wouldn’t worry about it for now

tjfulle:
On further thought, I don't think we should worry about assumed size arrays - intrinsic functions don't either.  For instance, the following will fail to compile

```
subroutine MySubroutine(x)
  real :: x(*), mysum
  mysum = sum(x)
end subroutine
```

But, if `mysum=sum(x)` is replaced with `mysum=sum(x(1:n))`, for some integer `n`, it will compile

PR #126, opened 2017-11-30
[WIP] doxy2swig based ForTrilinos documentation generation for Tpetra and Teuchos 

mattbement:


mattbement:
Updated some of the python scripts to those used in 12-12-branch, rather than 12-12-1, as they handle python 2/3 issues.

aprokop:
@mattbement I updated the PR: 
- Proper license on files
- More appropriate CMake logic

We cannot merge this PR until we full test it. This depends on sethrj/swig#46.

aprokop:
Per @mattbement:

> A quick note about the documentation...As you probably saw, Seth added the docstring output to swig.  However, doxygen's support for fortran leaves something to be desired.  It want types (which have previously been declared public) to have an additional public decorator in their description for it to pick up the documentation.  Since, swig is emitting valid fortran, Seth and both think that adding superfluous decorators is probably not the right thing.  I will spend a little time looking into what it would take to fix doxygen (or at least just submit a bug report).

> It looks like this was reported over 4 years ago. https://bugzilla.gnome.org/show_bug.cgi?id=706864

PR #141, opened 2017-12-05
Fix documentations and CMake extension

sethrj:


cmake/SwigModules.cmake
```diff
@@ -36,7 +36,7 @@ ENDIF()
 
 # Define extra output files
 set(SWIG_FORTRAN_EXTRA_FILE_EXTENSION "f90") # old CMake
-set(SWIG_FORTRAN_EXTRA_FILE_EXTENSIONS "f90")# new CMake
+set(SWIG_FORTRAN_EXTRA_FILE_EXTENSIONS ".f90")# new CMake
```
aprokop: I'm not sure about this yet. We need to really test and regenerate the source to make sure this does not affect it.

sethrj: Well, without this change, configuration fails on CMake 3.8

sethrj: I'm not sure why the CI failed -- it looks like all the unit tests passed but then at the end it says:
```
codecov_bash_uploader: line 1: syntax error near unexpected token `newline'
```

aprokop: > Well, without this change, configuration fails on CMake 3.8

Weird. I use CMake 3.10 and it works fine there.

aprokop:
> ./codecov_bash_uploader: line 1: syntax error near unexpected token `newline'

Lets rerun the tests.

cmake/SwigModules.cmake
sethrj: And you're in develop mode when it works?

aprokop: yes, always.

aprokop:
I just double checked and the clean build with this patch fails with CMake 3.10.0
```
/home/xap/code/trilinos/fortrilinos/packages/ForTrilinos/src/belos/test/test_belos_types.f90:19:8:

     use forbelos
        1
Fatal Error: Can't open module file ‘forbelos.mod’ for reading at (1): No such file or directory
```
The problem is I don't see it compiling `forbelos.f90` for example. This is with SWIG 1298f4ec4.

Reverting the patch, the code compiles and passes the tests.

@sethrj Can you update this PR to include only documentation PR? It currently seems that the CMake code related to Swig is highly dependent on the CMake version, and we need to figure that out.

aprokop:
I further tested and the patch is not need for CMake versions 3.3.1 and 3.8.2.
I'm closing as I cannot reproduce it. Feel free to reopen if you experience it again.

I'm going to cherry-pick the doc update from this branch with another PR.

PR #149, opened 2017-12-14
Increased Tpetra test coverage, updated test macros

tjfulle:
Tpetra: Adding tpetra_crsmatrix tests
Utils: Split testing macros up and fix bug

- Adding test coverage for TpetraCrsMatrix.  There are still many skeletons in
  the test file, but this is a good start.
- Broke the testing macros up in to two files.  The original,
  FortranTestMacros.h is essentially how it was before the FORTRILINOS_TEST_
  stuff was added.  All FORTRILINOS_TEST_ related macros were moved to their own
  file.  This file does not depend on DBCF and all line lengths should now be
  under 132.
- All `TEST_*` macros were made collective to prevent a test from hanging if one processor failed but others did not.  This may not be a good long term solution.


scripts/patches/apply-patches
```diff
@@ -10,10 +11,21 @@ if TRILINOS_DIR is None:
...
-this_dir = os.path.dirname(os.path.realpath(__file__))
-for filename in glob.glob(os.path.join(this_dir, '*.diff')):
-    diff_file = os.path.join(this_dir, filename)
-    print(diff_file)
-    command = 'patch -p 1 < {0}'.format(diff_file)
-    p = Popen(command, cwd=TRILINOS_DIR, shell=True)
+def apply_patch(patchfile, cwd=None):
+    command = 'patch -p 1 < {0}'.format(patchfile)
+    print(command)
+    return 0
```
aprokop: This must be a leftover from some debugging.

src/tpetra/test/test_tpetra_crsmatrix.f90
```diff
@@ -0,0 +1,864 @@
...
+  ADD_SUBTEST_AND_RUN(TpetraCrsMatrix_AlphaBetaMultiply)
+  ADD_SUBTEST_AND_RUN(TpetraCrsMatrix_ActiveFillGlobal)
+  ADD_SUBTEST_AND_RUN(TpetraCrsMatrix_SimpleEigTest)
+
+!  ADD_SUBTEST_AND_RUN(TpetraCrsMatrix_ActiveFillLocal)
+!  ADD_SUBTEST_AND_RUN(TpetraCrsMatrix_replaceColMap)
+!  ADD_SUBTEST_AND_RUN(TpetraCrsMatrix_replaceDomainMapAndImporter)
+!  ADD_SUBTEST_AND_RUN(TpetraCrsMatrix_removeEmptyProcessesInPlace)
+!  ADD_SUBTEST_AND_RUN(TpetraCrsMatrix_getCrsGraph)
+!  ADD_SUBTEST_AND_RUN(TpetraCrsMatrix_isStorageOptimized)
```
aprokop: Why are all these commented out? And where are the rest? Or is it just the first patch of many?

src/utils/src/fortest.f90
```diff
@@ -116,6 +114,47 @@ module fortest
 
   ! -------------------------------------------------------------------------- !
 
+  subroutine fortest_stop(message)
+    implicit none
+    character(len=*), intent(in), optional :: message
```
aprokop: I did not even know about this `optional` keyword. I wonder if it mean that instead of multiple functions for default values on the Fortran side we could have done optional arguments instead.

src/utils/src/fortest.f90
```diff
@@ -116,6 +114,47 @@ module fortest
...
+    integer, parameter :: stopcode=993299
```
aprokop: I assume this is just some random number?

scripts/patches/apply-patches
tjfulle: not necessarily, it is to give a visual of what is actually being done

aprokop: But you don't actually apply the patch?

src/tpetra/test/test_tpetra_crsmatrix.f90
tjfulle: These tests have not been completed.  I thought it would be good to get some test coverage in there.  I'll go back and pick off individual tests one at a time.

src/utils/src/fortest.f90
tjfulle: I believe it is an IEEE error code.  I got it from `DBCF.h`

src/utils/src/fortest.f90
tjfulle: Probably many of the functions could be consolidated.  Fortran also allows specifying arguments by the caller as keywords:

```
call the_subroutine(arg1, arg2=some_value)
```

tjfulle: The keyword form gives some flexibility in argument ordering by the caller when there are multiple optional arguments (you don't have to include optional arguments 3 and 4 if all you want is to send in 5)

scripts/patches/apply-patches
aprokop: What I mean is the `return 0` statment

src/utils/src/fortest.f90
aprokop: @sethrj Were you aware of this option? Or is it a SWIG thing that make multiple functions for default arguments? In that case, I wonder if that's configurable.

tjfulle:
@aprokop , the build is failing, but seems to be on the Trilinos side.  Unless I missed something...

scripts/patches/apply-patches
tjfulle: doh!  you are right

aprokop:
@tjfulle This is because a bug in `apply_patch`. The early return statement makes it skip patching which is required to be able to compile.

scripts/patches/apply-patches
tjfulle: good catch!

tjfulle:
@aprokop, oh, okay.  I put that script in but didn't know it was now being used by anyone else!

Issue #153, opened 2017-12-14
Hit 132 character limit


aprokop:
- [x] Teuchos
  - [x] Source
  - [x] Tests
- [x] Tpetra
  - [x] Source
  - [x] Tests
  - [x] Tutorial
    Do not use `EXPECT_TRUE` and other testing macros in tutorial, replace them by something a Fortran user will recognize.
- [ ] Simple
  - [x] Source
  - [ ] Tests
- [x] Belos
  - [x] Source
  - [x] Tests

tjfulle:
Tpetra tutorials are now from of `FortranTestMacros.h` (#168)

aprokop:
OK, the only thing remaining here is to update `simple` tests (similarly to tutorial) to fit into 132 char limit, and we will be able to compile with `flang`.

tjfulle:
I’m early done with ImportExport tests (they enemies up being more complicated than I originally thought).  When that is done, I can get the simple stuff done pretty quickly.  That will be the last remaining code using FortranTestMacros.h

PR #155, opened 2017-12-15
ForTeuchos: Use FortanTestingUtilities.h

tjfulle:
Use FORTRILINOS_UNIT_TEST for Teuchos.

Addresses: #153


tjfulle:
@aprokop, this only took a few minutes, feel free to reject.

PR #156, opened 2017-12-15
Script to auto generate unit tests

tjfulle:
@aprokop, this is the latest version of the script I use to parse swig fortran files and auto generate unit tests.  It is a total hack, but I almost accidentally erased it and thought I'd put it out for safe storage.  It might be useful for others too.

Feel free to reject.

### DIFF
--- a/scripts/autogen_tests.py
+++ b/scripts/autogen_tests.py
@@ -1,3 +1,7 @@
+#Copyright 2017, UT-Battelle, LLC
+#
+#SPDX-License-Identifier: BSD-3-Clause
+#License-Filename: LICENSE
 import re
 import sys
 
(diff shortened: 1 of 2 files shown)


scripts/autogen_tests.py
```diff
@@ -0,0 +1,224 @@
...
+        self.procs = procs
+        self.filename = filename
+
+    def write(self, stream):
+        if stream is None:
+            stream = sys.stdout
+
+        # Write the preamble to the test suite
+        stream.write("""\
+program test_{0}
```
aprokop: Could you please add the license block as the first few lines of the file?

tjfulle:
@aprokop, I'm not sure that it is feasible to compare a generated test file with an existing test file and write tests for the difference.  Once a file is generated, I have modified it so heavily that there wouldn't be any regular patterns to check against.

Issue #170, opened 2017-12-16
.neqv. broken in GCC 7.?


tjfulle:
@aprokop reported that

```
if (a .neqv. b) then
 ...
end if
```

gives erroneous results.  The culprit seems to be `.neqv.`.  I have not been able to reproduce the failures with GCC 6.4.

aprokop:
More background for the issue.

The comparison `a. eqv. b` and `a. neqv. b` fail when `a = c_associated(x)` and `b = true` even though both `a` and `b` are individually true. There is clearly something wrong as
```fortran
WRITE(*,"(L1,L2,L2,L2)") a, b, a.eqv.b, a.neqv.b
```
produces `T T F T` which does not make any sense.

Online search seems to pick up a compiler bug [issue](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=83254). It seems to affect gfortran 7.1 and 7.2. It is fixed upstream but has not been backported yet.

tjfulle:
Workaround provided in #171

PR #171, opened 2017-12-16
Workaround for .neqv. which does not seem to work with gnu 7.?

tjfulle:
@aprokop, this is a very simple workaround for the issues you are seeing with gnu 7.2 w.r.t. `.neqv.`.   I'm on my laptop at the moment and don't have access to 7.2, so I have only tested in 6.4, but this workaround avoids `.neqv.` altogether without changing user code.

Addresses: #170

### DIFF
--- a/src/teuchos/test/test_teuchos_comm.f90
+++ b/src/teuchos/test/test_teuchos_comm.f90
@@ -29,13 +29,13 @@ contains
 
 #ifdef HAVE_MPI
     call comm%create(MPI_COMM_WORLD); TEST_IERR()
-    TEST_EQUALITY_CONST(c_associated(comm%swigptr), .true.)
+    TEST_ASSERT(c_associated(comm%swigptr))
 
     call MPI_COMM_RANK(MPI_COMM_WORLD, comm_rank_f, ierr)
     call MPI_COMM_SIZE(MPI_COMM_WORLD, comm_size_f, ierr)
 #else
     call comm%create(); TEST_IERR()
-    TEST_EQUALITY_CONST(c_associated(comm%swigptr), .true.)
+    TEST_ASSERT(c_associated(comm%swigptr))
 
     comm_rank_f = 0
     comm_size_f = 1
--- a/src/teuchos/test/test_teuchos_plist.f90
+++ b/src/teuchos/test/test_teuchos_plist.f90
@@ -33,11 +33,11 @@ contains
     OUT0('Starting TeuchosPList_Basic!')
 
     call plist%create('myname'); TEST_IERR()
-    TEST_EQUALITY_CONST(c_associated(plist%swigptr), .true.)
+    TEST_ASSERT(c_associated(plist%swigptr))
 
     ! Test a function that raises an exception
     TEST_THROW(call load_from_xml(plist, 'nonexistent_path.xml'))
-    TEST_EQUALITY_CONST(c_associated(plist%swigptr), .true.)
+    TEST_ASSERT(c_associated(plist%swigptr))
 
     ! Get and set a vlaue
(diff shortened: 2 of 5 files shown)


aprokop:
@tjfulle It made one of the tests pass (`test_teuchos_comm`), but the second one (`test_teuchos_plist`) still fails:
```
Starting TeuchosPList_Basic!           
***ERROR: On Proc:   0                 
*** ERROR: File: 'test_teuchos_plist.f90', line     77                         
*** ERROR: TEST_EQUALITY_CONST(c_associated(sublist%swigptr),  .false.)        
Test TeuchosPList_Basic FAILED! 
```

tjfulle:
It should be easier now to fix the fortrilinos_eqv function now for edge cases and leave the rest the same

tjfulle:
@aprokop , when you get a chance, can you check the latest commit agains 7.2?

aprokop:
@tjfulle Nope, does not work:
```
write(*,*) a, b, a.eqv.b, a.neqv.b, bool_to_int(a), bool_to_int(b) 
T T F T           0           1 
```

tjfulle:
Both a and b are true?  And .eqv. Returns false and .neqv. True?

tjfulle:
How did this ever work before,  the most current attempt uses only .not. For comparison

tjfulle:
And bool_to_int returns 0 for a==.true.?  Is a the return value from c_associated?

aprokop:
So I just did some playing around, and it seems that the core problem is the fact that you used a function. Just the mere fact of passing it as a function argument changes something, I put ` write(*,*) a, .not.a, a.eqv. .true., a .neqv. .true., bool_to_int(a)` before the `_const` function call and inside, and it produces different results:
```
 T F T F           1                   
 T T T T           0
```
So, I think the fact that the previous macro did *not* use a function is important.

tjfulle:
I’m not sure how that would make a difference, but I can’t argue with the result.

Are the failing test with variables of type logical(c_bool) or just logical?

aprokop:
Plain logical.

tjfulle:
What if you just replace the call with `TEST_ASSERT((.not. <arg>))` for the false case and `TEST_ASSERT(<arg>)` for the true.  Does it pass or still fail?  If it passes, I’d be inclined to forget `TEST_EQUALITY_CONST` for now

aprokop:
I replaced `TEST_EQUALITY_CONST` with `TEST_ASSERT` only for the calls involving `c_associated`. This made the tests pass.

tjfulle:
Given that, I think we should just pitch `TEST_EQUALITY_CONST`.  It doesn’t add anything, and it is not like we need to replicate Teuchos.

tjfulle:
@aprokop , can you try one last thing before I pitch it?  Can you replace direct booleans in the call to TEST_EQUALITY_CONST with variables?  e.g., instead of `TEST_EQUALITY_CONST(<thing>, .true.)` do `true=.true.; TEST_EQUALITY_CONST(<thing>, true)`?

tjfulle:
you would have to have new variable names since `true` and `false` are all already taken for the `c_bool` variables

aprokop:
@tjfulle I tried that at some point and that did not work. The thing I remember working was something like `TEST_EQUALITY_CONST(logical(c_associated(...), kind=c_bool), logical(.true. kind=c_bool)`.

tjfulle:
Oh, that's ugly!  Latest push removes `TEST_EQUALITY_CONST`.  There are two places where i have `.eqv.`, in some test code, but one of your previous replies indicated that it may still work

aprokop:
I'm OK with merging this.

PR #188, opened 2017-12-19
develop promotion

aprokop:
Here, we generally would do app integration tests and/or more heavy testing. However, right now we are not integrated, so as long as develop CI is clean we can merge.


aprokop:
Merged manually to resolve conflicts.

PR #191, opened 2017-12-19
[DEVELOP BRANCH] Fix build badge in README.md

### LINKED ISSUE #190: Fix build badges issue
Right now, both build badges from both `master` and `develop`branches point to the same build. We need to separate them somehow. @dalg24 Do I need to create a new Jenkins build configuration for my master branch even if I want master and develop to do essentially the same CI testing? Or is there a way to avoid that?

dalg24:
Make it point to ForTrilinos-experimental instead of ForTrilinos-continuous.
Resolve #190.

NB: DO NOTE PROMOTE THIS COMMIT TO THE MASTER BRANCH

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 ForTrilinos
 ===========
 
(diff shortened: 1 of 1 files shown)


aprokop:
No, this is not we want.

PR #199, opened 2017-12-20
Allow ForTrilinos to build as an extra package in an extra repo

bartlettroscoe:
This adds the trivial files PackagesList.cmake and TPLsList.cmake that allows ForTrilinos to be built as an extra package in an extra repo when building with Trilinos.  So, instead of cloning ForTrilinos under Trilinos/packages/, you clone it under the base Trilinos/ directory and then configure Trilinos with

```
cmake -DTrilinos_EXTRA_REPOSITORIES=ForTrilinos -DTrilinos_ENABLE_ForTrilinos=ON \
  [other options] \
  ${TRILINOS_DIR}
```

In order to build ForTrilinos as part of Trilinos, you will need to do this since you can no longer build ForTrilinos as an inserted package with the Trilinos 'develop' branch (see the commits logged with trilinos/Trilinos#2085).

But note that adding these files does not stop ForTrilinos being treated as an inserted package with other versions of Trilinos (i.e. Trilinos 12.12).  These files will just be ignored in that case.


aprokop:
ok to test

aprokop:
This PR requires documentation update on RTD (and maybe other places).

bartlettroscoe:
Did not realize ForTrilinos had a develop/master workflow:

* https://docs.google.com/document/d/1uVQYI2cmNx09fDkHDA136yqDTqayhxqfvjFiuUue7wo/edit#heading=h.u2ougk1wk7ph

bartlettroscoe:
Some strange stuff is going on with the 'develop' and 'master' branches of this repo that caused a merge conflict of my topic branch.  Therefore, I cherry-picked the commits on top of my branch on top of 'develop' and no this looks okay.

aprokop:
@bartlettroscoe The strange stuff is being cause by conflicts in `README.md` which points to different Jenkins CI builds. I don't currently know how to properly resolve it (we need to different copies of files in `master` and `develop`).

aprokop:
@jwillenbring What is the cost of being an external package as opposed to inserted? Can it still be released as part of the trilinos tarball? In general, I feel like I have little understanding of pros and cons of each approach. If we become an external package, would it be difficult to change the type later?

aprokop:
@maherou We have an ongoing discussion about the type of package ForTrlinos is to Trilinos. I'd like to solicit your thoughts on the topic: from your point of view and your knowledge of the community, would you recommend releasing ForTrilinos as part of Trilinos (as a  regular package in the long term), or keeping it separated? My and @tjfulle's main concern is that it makes it even more painful for users to install it (Trilinos by itself is already painful, and this would require extra steps).

bartlettroscoe:
I am just saying that Trilinos should not include packages in an official release tarball that have no automated testing either before a release branch is created or after it is put out.

The issues if what the relationship is between Trilinos and ForTrilinos on their development branches is a different story but clearly what we were doing was not the right thing (i.e. defining a ForTrilinos package as an EX package so it is disabled by default but the Dependencies.cmake files still need to be without error which was not the case here).

tjfulle:
I would prefer to see ForTrilinos as a package within Trilinos and not an external package.  The ForTrilinos team could pretty quickly resolve any issues standing in the way.

aprokop:
OK, this is ready for merge now. But I'd like to keep the discussion going.

bartlettroscoe:
> OK, this is ready for merge now. But I'd like to keep the discussion going.

Note that you can actually do both but I think that would be confusing. The reason that I had to remove ForTrilinos as an inserted package as part of trilinos/Trilinos#2085 is that the repo is already cloned on a bunch of automated test machines and if that repo even exists, then the ForTrilinos package is automatically defined and its Dependencies.cmake files will be processed.   If we could scrub the ForTrilinos repo from all of the Trilinos clones on all of the test machines, then I could revert the commit trilinos/Trilinos@cae1480cc11448bcfabecb2edffaa81becfa5b1 and Trilinos automated builds wold automatically stop trying to process ForTrilinos.  But I don't know how to do that on all the test machines automatically.

An alternative would be to add a variable to Trilinos/PackagesList.cmake to optionally even list ForTrilinos or not.  That would also address the immediate problem, but it kind of hackish.

PR #201, opened 2017-12-20
Docker cleanup

aprokop:


aprokop:
@dalg24 Hmm, trapping more did not seem to help. According to [this](https://gist.github.com/datagrok/dfe9604cb907523f4a2f):
```
The default action for shell scripts upon receiving HUP, INT, and TERM is to abort, which triggers the 
EXIT handler just before execution ends. So for shell scripts, the EXIT trap is a good place for 
cleanup code.
```
I wonder why it does not clean up the containers then...

PR #203, opened 2018-01-10
Simple: added missing LINKER_LANGUAGE for tests

aprokop:


aprokop:
Thanks to @mattbement for finding this out.

aprokop:
The problem manifested only using Intel compilers, Flang and gcc managed to compile the code just fine.

PR #206, opened 2018-01-18
Remove free-length flags

aprokop:


sethrj:
I don't think this will ever work for our tests unless we ditch C preprocessing; it's impossible to avoid expansive macros (which are always rendered onto a single input line).

PR #207, opened 2018-01-28
WIP: SWIG memory and native fortran improvements 

sethrj:
Andrey,

This branch contains updated code for the new `develop` branch of SWIG. I've basically rewritten `ForTeuchos` for improved compatibility with Fortran:
- References now actually work; so you can do `unallocated_plist = get_sublist("foo")` and not get a crash
- All `std::string` const references are implicitly converted to and from Fortran character arrays; the `std::string` class is no longer even wrapped
- `ParameterList` returns `allocate`d arrays and strings instead of requiring the user presize them (or returning pointers).
- I've `%apply`ed native fortran integers to `size_t`, so users don't have to cast every argument to the longer C type.
- Logical types now use the native fortran components (and are much safer than before; apparently Fortran checks logicals by using the least significant bit `x & 1` rather than C's `x != 0`. So the value `-2` would be 'false' in fortran and 'true' in C.)

I've not updated any of the downstream classes. We can work together on this next week, but this should be a big improvement for intuitive Fortran usage of fortrilinos.


aprokop:
OK, I cherry-picked most of the code here with the exception of a) treating bool as int, and b) treating size_t as int. I think those warrant a separate discussion.

sethrj:
@aprokop the "bool as int" layer is necessary to guarantee consistency between fortran and C integer representations. The `size_t` vs native `int` definitely merits discussion but what's the problem with the `bool` wrapping?

sethrj:
@aprokop To bolster the argument for using the logical wrapper, I found this tidbit in the MPI Fortran wrapping code:
```
! However, a bunch of MPI Fortran subroutines use LOGICAL dummy
! parameters, and Fortran disallows passing LOGICAL parameters to
! BIND(C) routines (because the .TRUE. and .FALSE. values are not
! standardized (!)).  Hence, for these
! subroutines-with-LOGICAL-params, we have to be creative on how to
! invoke the back-end ompi_*_f() C function.
```

aprokop:
@sethrj Solid! So, what creative thing do they do?

sethrj:
See https://github.com/open-mpi/ompi/blob/master/ompi/mpi/fortran/use-mpi-f08/mpi-f-interfaces-bind.h ; it's rather lengthy and confusing because they have to deal with *arrays* of bools/logicals, but at its core they call a `OMPI_ARRAY_LOGICAL_2_INT` macro on it. It looks like they assume that a "true" value is a specific Fortran-compiler-defined value that they embed, using the configure-time https://github.com/open-mpi/ompi/blob/master/config/ompi_fortran_get_value_true.m4 file to determine it.

sethrj:
The `cray` compiler even provides a function to convert between C and logical values: http://docs.cray.com/books/S-2179-52/html-S-2179-52/ppgzmrwh.html#IYMRWH

sethrj:
https://software.intel.com/en-us/forums/intel-visual-fortran-compiler-for-windows/topic/275071#comment-1548435

https://software.intel.com/en-us/forums/intel-fortran-compiler-for-linux-and-mac-os-x/topic/529134

There are flags that can enforce a Fortran compiler to represent true/false as 1/0, but it's not required.

sethrj:
Superseded by #261

PR #208, opened 2018-01-29
Update swig

aprokop:
First part of @sethrj work. This only updates the wrappers with the most recent SWIG and fixes the build for that.

### DIFF
--- a/src/tpetra/test/test_tpetra_import_export.f90
+++ b/src/tpetra/test/test_tpetra_import_export.f90
@@ -124,8 +124,6 @@ contains
         call mine_parent%create(src, num_vecs+2)
         call neigh_parent%create(tgt, num_vecs+2)
         TEST_ASSERT((num_vecs == 5))
-        call mv_mine%create()
-        call mv_with_neighbors%create()
         mv_mine = mine_parent%subViewNonConst(cols)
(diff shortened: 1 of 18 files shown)


src/tpetra/test/test_tpetra_import_export.f90
```diff
@@ -170,7 +171,7 @@ program test_TpetraImportExport
       !                     [n-1  2n-1  3n-1  4n-1  5n-1]
       call mv_with_neighbors%doImport(mv_mine, importer, TpetraREPLACE)
       do j = 1, num_vecs
-        a = mv_with_neighbors%getData(j)
+        a => mv_with_neighbors%getData(j)
```
sethrj: Probably want to change the function name to something like `getView`

aprokop: The name is coming from Tpetra, don't necessarily want to change that.

sethrj: If we want to distinguish functions that return array pointers from functions that return allocated arrays, won't we have to change some function names?

tjfulle: FWIW, I might lean toward changing the name to `getView` to have consistency in ForTrilinos.  Or, keep `getData` and write a `getView` function that merely wraps `getData`.

PR #210, opened 2018-01-30
Introduce Teuchos::Array and update everything to use native strings

aprokop:


aprokop:
@tjfulle @bscollin  Can you please provide feedback on the last commit in this PR (fixing tests for flang)? I'm curious if you are familiar with the issue.

aprokop:
I opened a new issue #211 to keep track of the Flang problem.

PR #215, opened 2018-02-07
Teuchos::Comm improvements

sethrj:
- Simplify/improve typemap from the one I originally wrote
- Use ISO C compatibility layer for C/F type, and MPI types elsewhere
- Add MPI-fortran output of comm handle

### DIFF
--- a/src/teuchos/src/Teuchos_Comm.i
+++ b/src/teuchos/src/Teuchos_Comm.i
@@ -41,6 +41,12 @@ class Comm
 {
   public:
 
+    // Wrap built-in methods
+    int getRank() const;
+    int getSize() const;
+    void barrier() const;
+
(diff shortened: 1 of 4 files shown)


src/teuchos/src/Teuchos_Comm.i
```diff
@@ -41,31 +57,36 @@ class Comm
...
 
-    int getRank() const {
-      return $self->getRank();
-    }
-    int getSize() const {
-      return $self->getSize();
-    }
-    void barrier() const {
-      $self->barrier();
+    MPI_Comm getRawMpiComm() {
```
aprokop: Don't like this (as `getRawMpiComm` is really part of `MpiComm` and not `Comm`) but I understand the reason why it's here.

PR #216, opened 2018-02-12
[WIP] broke getGlobalRowView

tjfulle:


src/tpetra/test/test_tpetra_crsmatrix.f90
```diff
@@ -599,58 +603,135 @@ program test_TpetraCrsMatrix
...
+      TEST_ASSERT((.not. any(mask==-1)))
+      TEST_ARRAY_EQUALITY(cgview(mask), ginds)
+      TEST_FLOATING_ARRAY_EQUALITY(csview(mask), values, epsilon(zero))
+      deallocate(mask)
+
+      call A%fillComplete(params);
+
+      ! check for throws and no-throws/values
+      TEST_THROW(cgview => A%getGlobalRowIndicesView(gblrow))
+      TEST_THROW(csview => A%getGlobalRowValuesView(gblrow))
```
tjfulle: @aprokop, to be consistent with Tpetra, we could make a non-member `getGlobalRowView` procedure that mimics the C++ member version.  Something like:

```f90
subroutine getGlobalRowView(A, gblrow, cgview, csview)
   type(TpetraCrsMatrix), intent(in) :: A
   integer(global_ordinal_type), intent(in) :: gblrow
   real(scalar_type), dimension(:), pointer :: csview
   integer(global_ordinal_type), dimension(:), pointer :: cgview
  cgview => A%getGlobalRowIndicesView(gblrow)
  csview => A%getGlobalRowValuesView(gblrow)
end subroutine
```

aprokop: Hmm, ok. @sethrj Is it possible to inject code fragment into the Fortran wrapper for a specific class?

sethrj: Yes, there are several ways it could be done. If the function signature can be automatically converted correctly from C++ then it's best to use `%extend` on the class, then you can add code to the proxy function with `%fortranappend` or `%fortranprepend`. Or, you can replace the function call in the proxy code entirely using `%feature("shadow")`.

The *most* manual solution is to extend the fortran class by manually extending the proxy code. I think it would look like this: one fragment adds to the module code, the second injects code into the class type definition.
```swig
%extend TpetraCrsMatrix {
%proxycode {
subroutine tpetracrsmat_getGlobalRowView(A, gblrow, cgview, csview)
   type(TpetraCrsMatrix), intent(in) :: A
   integer(global_ordinal_type), intent(in) :: gblrow
   real(scalar_type), dimension(:), pointer :: csview
   integer(global_ordinal_type), dimension(:), pointer :: cgview
  cgview => A%getGlobalRowIndicesView(gblrow)
  csview => A%getGlobalRowValuesView(gblrow)
end subroutine
}
%insert("ftypes") {
getGlobalRowView => tpetracrsmat_getGlobalRowView
}
}
```

sethrj:
The workaround I posted is not valid with the latest SWIG versions.

I've updated the contents of @tjfulle's test and pushed to a new branch at https://github.com/sethrj/ForTrilinos/tree/fix-getGlobalRowView . It looks like the test now fails to compile, so that's a plus. I think it'll just be a simple matter of creating a new SWIGTYPE for "arrayview by reference".

sethrj:
Closing due to inactivity

Issue #177, opened 2017-12-17
Prefix enum names


aprokop:
Otherwise will certainly lead to conflicts. @sethrj sorry, only realized this point now. But we need a quick fix. Could we prefix the enums automatically? Ideally, I'd like to prefix them with something that depends on the module they are in.

aprokop:
Luckily, I think I [found](http://swig.10945.n7.nabble.com/How-to-strip-off-enum-prefix-td6373.html) a quick hack around the issue without touching SWIG. So, we can still keep chugging alone.

aprokop:
The workaround is merged as part of #178. But we need an automatic way of doing this (though this can be done after the release).

sethrj:
Done in #220 : before any of the `%include`s in each module, add
```swig
%rename("Teuchos%s",%$isenumitem) "";
%rename("Teuchos%s",%$isenum) "";
```

That adds a prefix to all the encountered enums. Adding a similar %rename directive for Tpetra after `%import <forteuchos.i>` does the same thing.

PR #220, opened 2018-02-27
Update SWIG modules to be compatible with the current SWIG

sethrj:
The most significant change was that it's no longer possible for `%fragment` to inject code when it's imported. This required a workaround for the exception handling code, and it also prevents the Tpetra types from propagating into downstream modules.


sethrj:
I've removed the warnings from the fortpetra and forteuchos modules. This PR also fixes #177  .

aprokop:
Superseded by #221.

Issue #219, opened 2018-02-27
Fix SWIG fortran extension 


sethrj:
On CMake 3.8 (and possibly other versions), the extension of the generated SWIG Fortran file is incorrect.  With the current version of the repo, `make VERBOSE=1` shows why the SWIG sources are already regenerated:
```
cd /Users/s3j/_code/_build/Trilinos-fortdev && /opt/local/bin/cmake -E cmake_depends "Unix Makefiles" /usr/local/src/Trilinos-public /usr/local/src/Trilinos-public/packages/ForTrilinos/src/utils/src /Users/s3j/_code/_build/Trilinos-fortdev /Users/s3j/_code/_build/Trilinos-fortdev/packages/ForTrilinos/src/utils/src /Users/s3j/_code/_build/Trilinos-fortdev/packages/ForTrilinos/src/utils/src/CMakeFiles/forerror.dir/DependInfo.cmake --color=
Deleting primary custom command output "/usr/local/src/Trilinos-public/packages/ForTrilinos/src/utils/src/forerrorFORTRAN_wrap.cxx" because another output "/usr/local/src/Trilinos-public/packages/ForTrilinos/src/utils/src/forerrorf90" does not exist.
```
Note that the file it looks for is `forerrorf90` not `forerror.f90`.

If the existing file `forerror.f90` in the repository is deleted, then the build fails:
```
Deleting primary custom command output "/usr/local/src/Trilinos-public/packages/ForTrilinos/src/utils/src/forerrorFORTRAN_wrap.cxx" because another output "/usr/local/src/Trilinos-public/packages/ForTrilinos/src/utils/src/forerrorf90" does not exist.
```

See !141

The fix is to apply the following patch
```
$ git show e54811f2d7fb956776954ab9d9a1be89febc9925
commit e54811f2d7fb956776954ab9d9a1be89febc9925
Author: Seth R Johnson <johnsonsr@ornl.gov>
Date:   Tue Dec 5 15:45:19 2017 -0500

    Fix SWIG CMake code for newer CMake versions

diff --git a/cmake/SwigModules.cmake b/cmake/SwigModules.cmake
index 678509c..c357e60 100644
--- a/cmake/SwigModules.cmake
+++ b/cmake/SwigModules.cmake
@@ -36,7 +36,7 @@ ENDIF()

 # Define extra output files
 set(SWIG_FORTRAN_EXTRA_FILE_EXTENSION "f90") # old CMake
[...]

PR #217, opened 2018-02-20
[WIP] native Fortran pointer conversion to/from arrayview

sethrj:
This allows `Teuchos::ArrayView<T>` to replace current uses of `std::pair<T*, size_t>`.

WIP because the code is untested because I'm working on a different branch of SWIG at the moment. If you'd like to make sure the code works, go ahead and work on this branch, or let me know if you want to wait and I will test/integrate later.

Closes #163

Issue #185, opened 2017-12-18
Eliminate all warnings


aprokop:

tjfulle:
It’s really too bad any meaningful warnings get lost In a sea of teuchos warnings...

aprokop:
I actually don't see any Teuchos warnings from our builds, for instance [here](https://cloud.cees.ornl.gov/jenkins-ci/job/ForTrilinos-continuous/72/warnings29Result/).

tjfulle:
On my machine I have a sea of warnings about throwing from destructor.  How do I rid myself of those?

aprokop:
> On my machine I have a sea of warnings about throwing from destructor. How do I rid myself of those?

It could be that the reason we don't see them at the moment is due to not enabling a lot of flags. I don't know of a robust solution do disable warnings only for some Trilinos packages, which is super annoying.

aprokop:
> On my machine I have a sea of warnings about throwing from destructor. How do I rid myself of those?

Apparently, the only thing you needed to do is to post to ForTrilinos about it, and now they're [being addressed](https://github.com/trilinos/Trilinos/pull/2089).

sethrj:
I've eliminated the remaining SWIG warnings in #220 .

PR #221, opened 2018-03-21
WIP: Upgrade ForTrilinos and enable TpetraOperator derived type

### LINKED ISSUE #163: Implement automatic converters for ArrayView and ArrayRCP
Right now, I put a bunch of manual wrapper functions for those implementing essentially the same thing. Need to alias `ArrayView<T>` to `std::pair<T*,size_t>` which should reduce the amount of code.

sethrj:
This is an initial implementation of the callback functionality for the Tpetra operator. I've included a skeleton for a unit test that demonstrates its use. This MR supersedes #220 , most of #207 (aside from the int/logical type conversions), and the largest parts of #217.

Closes #163, #177, #185, #219

### DIFF
--- a/src/teuchos/src/Teuchos_ArrayRCP.i
+++ b/src/teuchos/src/Teuchos_ArrayRCP.i
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2017, UT-Battelle, LLC
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ * License-Filename: LICENSE
+ */
+%{
+#include "Teuchos_ArrayRCP.hpp"
+%}
+
+%include <view.i>
+
+namespace Teuchos
+{
+template<typename T>
+class ArrayRCP
+{
+  public:
+    typedef T value_type;
+    typedef std::size_t size_type;
+};
+}
+
+// Create RCP with no-ownership flag (it's set to false)
+%define TEUCHOS_ARRAYRCP_IMPL(CONST, CTYPE)
+  // Treat const references as values
+  %apply Teuchos::ArrayRCP<CONST CTYPE> { const Teuchos::ArrayRCP<CONST CTYPE>& }
+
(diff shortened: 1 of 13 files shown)


sethrj:
@aprokop To get the builds to pass, unfortunately you'll have to remove the `-std=f2003` flag:
```

   call c_f_pointer(cptr=fself_ptr, fptr=handle)
                                        1
Error: TS 29113/TS 18508: Noninteroperable array FPTR at (1) to C_F_POINTER: Expression is a noninteroperable derived type
/scratch/source/trilinos/release/packages/ForTrilinos/src/tpetra/src/fortpetra.f90:6323:38:

 call c_f_pointer(cptr=fself_ptr, fptr=handle)
                                      1
Error: TS 29113/TS 18508: Noninteroperable array FPTR at (1) to C_F_POINTER: Expression is a noninteroperable derived type
```
This error is bogus, and is being fixed in the latest GCC, https://gcc.gnu.org/bugzilla/show_bug.cgi?id=84924 . But it'll be a while before any compiler sees that fix.

aprokop:
Hmm, it explains the gcc failure. Flang compiles fine, but the failure of the test is probably due to missing functionality.

aprokop:
I'd like to keep the flag to make sure we don't accidentally use any Fortran 2008 features. Ideally, the gcc folks would backport the patch into older compiler versions. But a) it unclear whether that would be done, and b) it may take a long time to get it to Ubuntu. I could switch to one of the recent gcc's once the patch is upstream. Should be easy with Spack. In the mean time, I probably need to follow @sethrj' advice and disable the flag. If I recall correctly, Flang just outright ignores that flag anyway.

sethrj:
FWIW I'm continuing to build the SWIG test suite with `-std=f2003` so we should have at least some assurance of no F2008 features with that.

sethrj:
All the useful parts of this have been incorporated into master.

PR #226, opened 2018-04-04
Inversion-of-control for linear operators

aprokop:
The main problem right now is that the test does not pass when running in parallel. For some reason, the Y vector coming from Belos gets NaN after the first couple iterations. This does not happen in serial, and does not happen in C++ driver in parallel with the same matrix and vectors.


sethrj:
Ew. I hope it's not something wacky with Fortran pointers and MPI (though I can't imagine why it would be). Have you added code to print the operator C addresses whenever they're created, being called, etc?

aprokop:
While the situation is still not 100% clear, some investigation seems to indicate that the vector Y coming from Belos may be unitialized, and can potentially contain denormalized or signaling NaN entries. This may be the case only for $beta = 0$. The workaround is to treat that case specially, and instead of doing $Y = 0*Y$, so $Y = 0$ first.

PR #235, opened 2018-06-07
added fortest.mod to install

mattbement:
To get an external build of the simple_eigen_handle test to work, this module file needs to be in the installed include directory.


aprokop:
@mattbement Why do we need to install `fortest.mod`? It should only be used internally.

mattbement:
Dunno - thought it was needed for comparisons of fortrilinos vectors/matrices etc.

aprokop:
What do you mean for comparisons? In `fortest` module, we have setting up and tearing up modules, and some routines to compare results in Fortran arrays. This is not a general Trilinos module, only stuff required for tests.

Do you have a use case where a user needs that?

mattbement:
I guess not - I was just thinking it contained an easy way to compare two matrices/vectors, which would be useful, but I didn't look in the module.

aprokop:
@mattbement No. Comparing vectors needs to compute the norm of the difference. Comparing matrices is not really possible unless you fetch internal arrays.

PR #241, opened 2018-09-14
[WIP] turned a couple functions into subroutines to sidestep an intel compiler issue

mattbement:
With these mods, all tests pass on the intel compiler.


aprokop:
Unfortunately, this is opposite to the principle of maintaining the original Trilinos function signatures. Do we have access to anybody from intel compilers team?

sethrj:
Closing due to inactivity

PR #243, opened 2018-09-19
Allow computation for complex eigenvalues

aprokop:
This changes API:
- Introduces `eIndex` (from Anasazi)
- Stores real and complex parts of eigenvalues

@amklinv @mattbement This may affect your work as it changes the API for the eigensystem. I still don't think the interface is final, and would like to have something better in the future. But don't have time to work on it right now. Any suggestions are welcome.

### DIFF
--- a/src/simple/src/eigen_handle.cpp
+++ b/src/simple/src/eigen_handle.cpp
@@ -105,8 +105,10 @@ namespace ForTrilinos {
       }
     }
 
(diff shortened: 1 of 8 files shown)


amklinv:
I'm not currently using the eigensolvers through ForTrilinos, so this does not impact my work.  Thanks for checking though!

mattbement:
No objections - we definitely want to support complex eigenvalues.

PR #251, opened 2018-12-23
Update Docker container

aprokop:
- Update flang version
  Requires gcc-7 build now and using gawk instead of mawk
- Update spack to v0.12.0
- Update cmake to 3.10.3
  Trilinos develop requires 3.10+ now


aprokop:
The new image has already been uploaded.

PR #254, opened 2018-12-26
Update testing

aprokop:


aprokop:
After this PR is merged, will need to check whether continuous testing would build against `develop` Trilinos instead of the one in the Docker container.

PR #260, opened 2019-01-30
Update to the latest SWIG

aprokop:
SWIG hash: 01c436496


src/tpetra/src/fortpetra.i
```diff
@@ -32,11 +32,11 @@ typedef char                                    Packet;
 %}
 
 // FIXME: Restore previous bool behaviour
-FORT_FUND_TYPEMAP(bool, "logical(C_BOOL)")
+FORT_FUND_TYPEMAP(bool, logical(C_BOOL))
```
sethrj: The FORT_FUND_TYPEMAP will have to change when the pending PR on `swig-fortran` is merged.

PR #261, opened 2019-02-04
Simplify and improve wrappers

### LINKED ISSUE #127: Running `make` second time in DeveloperMode causes recompilation of some code
Need to figure out why.

### LINKED ISSUE #211: Assignment to unallocated arrays fails in flang
```Fortran
   ! FIXME: without these allocation, the test segfaults when compiled with flang
    ! According to [this](https://stackoverflow.com/questions/42140832/automatic-array-allocation-upon-assignment-in-fortran) it
    ! seems that it should be supported in F2003. So it is not clear why flang fails.
    allocate(iarr(10))
    allocate(darr(10))
    iarr = plist%get_arr_integer('intarr')
```

### LINKED ISSUE #250: Change f90 to F90
Related to sethrj/swig#83.

sethrj:
Infrastructure changes:
- Update SWIG scripts for CMake 3.10 and higher (should still be backward compatible)
- Use `.F90` instead of `.f90` so that the compiler preprocesses automatically (newer CMake can produce a bunch of bogus warnings because it tries to apply the `-cpp` flag to already-preprocessed code)

- Wrap local ordinals with native fortran `int`
- Wrap booleans with native fortran `logical`
- Add typemaps for `ArrayView` to delete a lot of custom `%extend` code
- Convert some manual input transformations (e.g. subtracting 1) to typemaps
- Remove wrapping for deprecated Tpetra functions
- Clean warnings in unit tests (see #185)

- Closes #250 
- Closes #211 because we're showing that the interface returns a pointer; we would rarely need to allocate a result implicitly from that returned pointer
- Closes half of #163 : I haven't implemented the ArrayRCP yet but will soon
- Closes #127 for new versions of CMake, and I don't think we need worry about older verisons


sethrj:
@aprokop Can you direct me how to see the test messages (and possibly the Flang backtrace) from the failing tests? They all pass on my machine...

aprokop:
The latest build (can you not access the "Details" about to see Jenkins PR?) fails in both GCC and Flang with SegFault in `teuchos_plist` test. No backtrace is available. You (or I) can reproduce that by firing up a Docker container and running the build from there.

sethrj:
Yeah I saw the "details" listing the failing tests, but I couldn't see what it reported as the failure. I don't have Docker set up on my machine, perhaps we could get together sometime and debug these failing tests on yours?

aprokop:
@sethrj oh, boy, you've been prolific! Will review this today.

sethrj:
@aprokop Sweet. I also just pushed a fix for https://github.com/trilinos/ForTrilinos/pull/216 that includes some basic ArrayRCP wrapping, simplifying the manual code even further.

sethrj:
@aprokop I think the changes in this branch might have fixed some of the crashes @mattbement (I think?) was seeing on Titan. At least, it'd be good to take another shot at running previously failing tests with this branch.

sethrj:
@aprokop I had hoped that the maintenance gain of not having to duplicate a bunch of Tpetra constructor declarations in SWIG would outweigh the penalty of having a few useless (Kokkos node) constructors. 😔

I have previously thought about implementing a feature that would allow particular classes/types to cause declarations where they're used (such as constructors) to be ignored automatically. I think I have a means of doing this, but it's not typical/standard SWIG practice to do. That hasn't stopped us before though 😉

tjfulle:
It may be worth noting that Trilinos seems to have (finally) adopted (re-adopted?) a release cycle.  Tpetra is pushing through a bunch of deprecations now so that code can be removed this spring summer.  Things on the docket for removal are: 
- template arguments other than `Scalar`
- `DynamicProfile` - only `StaticProfile` `CrsMatrix` and `CrsGraph` will be available after April 16th (or later)

Can anything be done now to create Tpetra wrappers without the template arguments?

aprokop:
>     * `DynamicProfile` - only `StaticProfile` `CrsMatrix` and `CrsGraph` will be available after April 16th (or later)

@tjfulle You mean `DynamicProfile` would just be deprecated, right?

> Can anything be done now to create Tpetra wrappers without the template arguments?

Not sure what you mean here. Right now we hardcode the template arguments. If Tpetra leaves just Scalar, we still need to pass that through, right?

tjfulle:
@aprokop `DynamicProfile` will be marked deprecated on the minor release made on/around April 16.  We plan to have `DynamicProfile` removed from the code by the next major release.  I'm not certain exactly how that will be done, but it is possible that we will start removing code from the develop branch shortly thereafter.

sethrj:
@aprokop I think I've addressed all the feedback: everything that had Kokkos types is now gone.

As to the remaining question about `Teuchos::Array`: the constructor allows creation of an Array from a fortran array pointer.

aprokop:
So the `%fortranonlywrapped` is the new keyword to skip things? Cool! It is off by default to follow usual SWIG approach?

sethrj:
Yep!

Issue #136, opened 2017-12-02
Tpetra{Export,Import} test coverage 


aprokop:

tjfulle:
@aprokop , I have the tests done.  But, to properly test `import` and `export`, the `doImport` and `doExport` methods of `TpetraMultivector`, `TpetraCrsGraph`, and `TpetraCrsMatrix` need to be implemented.  `TpetraMultiVector` is the priority to push the current tests.

aprokop:
Great.

Yes, I've dragging my feet on implementing those. Can't do them automatically (as they belong to the base class), but can do semi-automatically and that should not be hard. Are `doImport` and `doExport` of the classes you listed the only things you need?

tjfulle:
As of right now, yes.  The tests I have written are invalid until there is actually something imported/exported to check :)

tjfulle:
The tests only use `TpetraMultiVector`

aprokop:
I have a patch that does `doImport` and `doExport`, but unfortunately, it does not compile right now due to the clash in `zero`: some tests define zero, and another zero is coming from Tpetra's CombineMode.

aprokop:
Now that #178 is merged, I hope that's sufficient for you to finish the tests.

aprokop:
Fixed in #180.

aprokop:
Actually, lets keep it open until we get 100% coverage of the classes.

aprokop:
Fixed by #262.

aprokop:
Fixed by #262.

Issue #135, opened 2017-12-02
TpetraCrsGraph test coverage 


aprokop:

kruger:
This is being addressed by PR #262.

Notes on minor issues with latest push (as of 04-24-2019).

Tests that do not work:
```
  !TODO-FIX ADD_SUBTEST_AND_RUN(TpetraCrsGraph_isIdenticalTo)
  !TODO-FIX ADD_SUBTEST_AND_RUN(TpetraCrsGraph_replaceDomainMapAndImporter)
```

Tests that need improvement
```
    fresult = Graph1%isIdenticalTo(Graph2)
    fresult = Graph%getImporter(); TEST_IERR()
    fresult = Graph%getExporter(); TEST_IERR()
    fresult = Graph%description(); TEST_IERR()
    call Graph1%swap(Graph2); TEST_IERR()
```

Tests that fail if TEST_THROW is used
```
    !TEST_THROW(call graph%setParameterList(fresult))
    !TEST_THROW(call Graph%getGlobalRowCopy(globalrow, indices(:), numindices))
    !TEST_THROW(call Graph%getLocalRowCopy(localrow, indices(:), numindices))
    !TEST_THROW(call Graph%replaceColMap(newcolmap))
    !TEST_THROW(call Graph%removeEmptyProcessesInPlace(Map))
    !TEST_THROW(call Graph%computeGlobalConstants(farg))
```

To see the problems with `TEST_THROW` just uncomment out one of the above methods.

aprokop:
First of all, there was a problem in some of the macros, producing wrong messages like `Test NAME FAILED!` without the proper name. I pushed the fix into the brunch. 

Second, I started looking at the `TEST_THROW` issues that you posted, and I think they are actually valid failures:
- `TEST_THROW(call Graph%getGlobalRowCopy(globalrow, indices(:), numindices))`
  This surprised me too. Apparently, it returns an empty row even if global index is invalid. It's undocumented in Doxygen, but is in the code.
- `TEST_THROW(call Graph%getLocalRowCopy(localrow, indices(:), numindices))`
  same as above

The rest `TEST_THROW` I'm not sure why you think they should throw.

kruger:
I can't get your recent changes to compile.  I get errors with both flang on the docker image and gfortran.  They both give errors of the form
```
/scr_phi/kruger/triroot/trilinos/packages/ForTrilinos/src/teuchos/test/test_teuchos_comm.F90:24:88:

[...]

Issue #130, opened 2017-12-02
Test eigen handle interface for more than one eigenvalue


aprokop:

aprokop:
The easiest way to test this would be with MPACT, so lets wait till #161 is done.

aprokop:
See #218.

kruger:
In PR #262  this is handled by the test_eigen_unit test.  The only catch is that the "2nd eigenvalue" is actually zero -- the matrix has only 1 non-zero eigenvalue, 3, which is also tested.  I think this is sufficient to close this issue.

Issue #120, opened 2017-11-29
TpetraCrsMatrix test coverage


tjfulle:

tjfulle:
@aprokop, this test is ready, minus one addition to `fortpetra.f90` that would be very useful.  That is, an overload of `TpetraCrsMatrix%insertGlobalValues` with the following signature:

```
subroutine insertGlobalValues(this, globalrow, cols, vals)
  class(TpetraCrsMatrix) :: self
  integer(C_LONG_LONG), intent(in) :: globalrow
  integer(C_LONG_LONG), dimension(:) :: cols
  real(C_DOUBLE), dimension(:) :: vals
  ...
end subroutine
```

This would call the same `swigc_` procedure as the version with `insertGlobalValues(this, globalrow, nument, vals, inds)`, but would infer `nument` by `nument = size(cols)`.  

And similarly for `sumIntoGlobalValues`, `replaceGlobalValues`.  

I implemented all of these methods before, you can see them [here](https://github.com/trilinos/ForTrilinos/pull/112/files)

aprokop:
@tjfulle OK, I'll take a look. I would prefer to first merge the MultiVector test, though.

tjfulle:
@aprokop said:

> I would prefer to first merge the MultiVector test, though

See #122

tjfulle:
@aprokop , I just pushed `test_tpetra_crsmatrix.f90`.  It does not compile right now, but will when the `TODO`s are resolved.  It compiled with a previous version of `fortpetra.f90` that I hand modified to include a few different `TpetraCrsMatrix` function overloads.  Once the necessary modifications are done in swig, it will compile.  There are some tests that don't pass, indicating some issues in `fortpetra.f90`.  They are also documented with `TODO`s (sorry, should have marked them `FIXME`).  The commit can be seen at 1c12e475f82cab8ba541b2f55085cc8832515ad6

tjfulle:
Also, @aprokop, if you are confused what the `TODO: Remove`, `TODO: Reinstate` tags mean - the tests from Tpetra from which I copied these tests dealt only with `*LocalValues` methods, but these methods were not available in `fortpetra`, so I used the `*GlobalValues` version.  I'd like to remove the `*GlobalValues` versions in favor of the `*LocalValues` to have parity with the Tpetra tests.

aprokop:
Fixed by #262.

PR #262, opened 2019-03-18
Improve test coverage of CrsGraph and CrsMatrix

kruger:
This goes through and increases coverage in CrsGraph and CrsMatrix.
Many of the methods are just testing that they run and not that they run
correctly because they use this idiom:
    fresult = graph%getFoo(); TEST_IERR()
This probably needs improving.

gaussSeidel is an example of a test that needs further work.

Also, the skeleton for these tests was done awhile ago and the API has
changed.  Coverage tests should be used to see what is missing once the
nightlies come through.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
-build
+build*
 *_autogen.f90
 *~
+.*.swp
+docker/.env
+TriBITS
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,6 @@
 ##---------------------------------------------------------------------------##
 
 TRIBITS_PACKAGE_DECL(ForTrilinos)
-set(ForTrilinos_PROJECT_SRCDIR ${CMAKE_CURRENT_SOURCE_DIR})
 
 ##---------------------------------------------------------------------------##
 ## Set up package-specific options
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,7 @@
 ##---------------------------------------------------------------------------##
 
 TRIBITS_PACKAGE_DECL(ForTrilinos)
+set(ForTrilinos_PROJECT_SRCDIR ${CMAKE_CURRENT_SOURCE_DIR})
 
 ##---------------------------------------------------------------------------##
 ## Set up package-specific options
--- a/cmake/CheckSWIGFortran.cmake
+++ b/cmake/CheckSWIGFortran.cmake
@@ -13,8 +13,6 @@ if (NOT SWIG_CHECKED_FORTRAN)
     OUTPUT_VARIABLE SWIG_help_output
     ERROR_VARIABLE SWIG_help_output
     RESULT_VARIABLE SWIG_help_result)
-  MESSAGE(STATUS "SWIG_help_result: ${SWIG_help_result}")
-  MESSAGE(STATUS "SWIG_help_output: ${SWIG_help_output}")
   if (NOT SWIG_help_output MATCHES "Fortran")
     message(FATAL_ERROR "This version of SWIG does not support Fortran "
       "wrapping. Please install the version from "
--- a/cmake/CheckSWIGFortran.cmake
+++ b/cmake/CheckSWIGFortran.cmake
@@ -13,6 +13,8 @@ if (NOT SWIG_CHECKED_FORTRAN)
     OUTPUT_VARIABLE SWIG_help_output
     ERROR_VARIABLE SWIG_help_output
     RESULT_VARIABLE SWIG_help_result)
+  MESSAGE(STATUS "SWIG_help_result: ${SWIG_help_result}")
+  MESSAGE(STATUS "SWIG_help_output: ${SWIG_help_output}")
   if (NOT SWIG_help_output MATCHES "Fortran")
     message(FATAL_ERROR "This version of SWIG does not support Fortran "
       "wrapping. Please install the version from "
--- a/docs/source/developer_tools.rst
+++ b/docs/source/developer_tools.rst
@@ -31,12 +31,15 @@ Then to launch an interactive Bash session inside that container, do:
 
     [host]$ docker-compose exec fortrilinos_dev bash
 
-Patch, configure, build, and test as you would usually do:
+The docker container supports building and testing with both gfortran and flang.
+Patch, configure, build, and test as you would usually do (ignore the flang step
+if using gfortran):
 
 .. code:: bash
 
     [container]$ cd $TRILINOS_DIR/ForTrilinos
     [container]$ mkdir build && cd build
+    [container]$ source ../scripts/docker_flang70_env.sh /scratch/spack # if flang
     [container]$ ../scripts/docker_cmake
     [container]$ make -j<N>
     [container]$ ctest -j<N>
@@ -78,5 +81,5 @@ This can be done by adding the following configuration options to the script (as
 
 
 .. .. warning::
-.. 
+..
     .. When using the developer mode, all of the patches ``scripts/patches`` directory of the ForTrilinos source tree must be applied.  If you previously used the ``scripts/patches/apply-patches`` script, this was done for you. Otherwise, be sure that all of the patches are applied.
--- a/scripts/autogen_tests.py
+++ b/scripts/autogen_tests.py
@@ -215,10 +215,10 @@ def main():
         if proc in ('create', 'release'):
             continue
 
-        if proc.startswith(('set_', 'get_')):
-            # Not sure about RowInfo right now
-            assert 'RowInfo' in proc2
-            continue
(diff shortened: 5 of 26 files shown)


aprokop:
ok to test

kruger:
Is there an easy way of seeing the configure scripts for the Jenkins jobs?  I found the CMakeCache.txt, but if there is a script, that would be easier (and maybe I am just missing something).

aprokop:
@kruger yes, the scripts are in `scripts` directory and are named `docker_cmake*`. There are several of them, depending on configuration.

kruger:
This PR addresses Issue #135 and some of the more detailed issues are there as well.

kruger:
The jenkins test result is similar to my own tests:  `graph1%isIdenticalTo(graph2)` is failing.   I welcome suggestions on how to fix this.

aprokop:
@kruger Interesting, just realized that the single coverage post here is *updated* by the coverage. So, in fact, we have the coverage report from the latest build.

kruger:
@aprokop   Interesting.  Any idea of what the fortpetra.f90 is below 50%?  If I read it correctly, it says that this branch improved things by 87%, but I can't see how to tell what's missing in coverage.

aprokop:
@kruger Take a look [here](https://codecov.io/gh/trilinos/ForTrilinos/src/increase-cov/src/tpetra/src/swig/fortpetra.F90)

kruger:
Thanks!  
The main issue is that not all permutations of a given method are tested.  I propose that we finish the PR, close the related issue, and then open a new issue to deal with it.  This can be done when funding is sorted.  The issue will be to increase coverage by increasing the ways of calling a method.   Maps and MultiVectors are in worse shape than CrsMatrix so this would allow a uniform approach to cleaning them all up.

As a comment for future reference:  I think the main thing for maps/matrix/graphs is more 2D decomposition (or perhaps, any 2D decomposition).  I have a note in the helper functions about this probably being a good idea.  If we create a second version of the helper routines with the 2D decomp and then program up the same type of tests, I think it would be easy to get most of the missing coverage. 

There are some miscellaneous helper routines and if/else branches that we are missing as well, but these are relatively minor.

aprokop:
> The main issue is that not all permutations of a given method are tested. I propose that we finish the PR, close the related issue, and then open a new issue to deal with it. This can be done when funding is sorted.

Agreed

> The issue will be to increase coverage by increasing the ways of calling a method. Maps and MultiVectors are in worse shape than CrsMatrix so this would allow a uniform approach to cleaning them all up.

Yes, Maps and MultiVectors would be of higher priority.

> As a comment for future reference: I think the main thing for maps/matrix/graphs is more 2D decomposition (or perhaps, any 2D decomposition).

I don't understand what 2D you are talking about. I think for me 2D may mean something different. Can you please elaborate?

> There are some miscellaneous helper routines and if/else branches that we are missing as well, but these are relatively minor.

I have rebased the PR. Now I need to look through it, and see if everything looks OK. Once I've done that, I will merge it.

kruger:
Maybe I misunderstood the tpetra tutorial, but I thought if you specify just a row map for the matrix constructor, then you have just the row disributed where as if you specify both row and column, then you have both distributed.   In any case, we need permutations on the constructors primarily if I am understanding how the swig wrappers work.

aprokop:
@kruger I rebased this PR on top of current master. Couple things that I had to fix were:
- Replaced all DynamicProfile with StaticProfile
  DynamicProfile is deprecated, and will be gone in the future
- Removed getLocalDiagOffsets tests
  @sethrj discovered issues with the function, and commented it out in his earlier PR. 

Please take a look, and I think it would be nice to do some more thorough testing in some functions. `TEST_NOTHROW` only tests that the call does not throw, not that it is produces the correct results. I think the higher priority should be placed on functions that could potentially deal with plus-minus one issues.

aprokop:
There is some problem in compilation with `test_tpetra_crsmatrix_helper`. I am able to reproduce it locally, so it's not spurious.

aprokop:
I wonder if it is some kind of race condition. You try to compile that source file multiple times in that directory. If different make threads will try to compile it simultaneously, it is possible that one of them is recompiling, while the second one is trying to use it. Something like this?

aprokop:
I think one way to fix that would be to do a dummy library for all `test_tpetra_*.F90` files, and then do `DEPLIBS` in `TRILINOS_ADD_EXECUTABLE_AND_TEST`.

kruger:
This was PR was originally meant to cover only Issue #135, but I think we can close that issue as well as Issue #120, Issue #121, Issue #122, Issue #130, and Issue #136.

aprokop:
Hmm, I've got 2 failure locally that I need to take a look at (at least, in Debug mode).
- `test_eigen_unit`
  ``` 
  At line 212 of file /home/xap/code/trilinos/fortrilinos/ForTrilinos/src/interface/test/test_eigen_unit.F90                                                                                     
Fortran runtime error: Index '0' of dimension 1 of array 'evalues' below lower bound of 1
  ```
  This is obviously a mistake in the indexing
- `test_tpetra_multivector`
  Some segfaults that warrant investigation

I have made some changes to the PR, and was not careful to stack them on top of your commit, instead amending it. I cannot force-push to the branch. Can you please add me as a collaborator for the TechX ForTrilinos fork?

aprokop:
OK, I pushed. Go ahead and fix the `test_eigen_unit` test. I'm looking into the multivector one.

kruger:
@aprokop   I'm not seeing the multivector problem.  I'm using `docker_cmake` which is in Debug mode.  What are you seeing?

aprokop:
I see things like this:
```
Thread 1 "ForTrilinosTpet" received signal SIGSEGV, Segmentation fault.
0x00007ffff769b87a in Teuchos::RCP<Tpetra::MultiVector<double, int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace> > >::assert_valid_ptr (this=0x4010000000000000) at /home/xap/code/trilinos/fortrilinos/packages/teuchos/core/src/Teuchos_RCP.hpp:540
540       if (ptr_)
(gdb) bt
#0  0x00007ffff769b87a in Teuchos::RCP<Tpetra::MultiVector<double, int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace> > >::assert_valid_ptr (this=0x4010000000000000) at /home/xap/code/trilinos/fortrilinos/packages/teuchos/core/src/Teuchos_RCP.hpp:540
#1  0x00007ffff7685894 in Teuchos::RCP<Tpetra::MultiVector<double, int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace> > >::debug_assert_valid_ptr (this=0x4010000000000000) at /home/xap/code/trilinos/fortrilinos/packages/teuchos/core/src/Teuchos_RCPDecl.hpp:832
#2  0x00007ffff767ac2c in Teuchos::RCP<Tpetra::MultiVector<double, int, long long, Kokkos::Compat::KokkosDeviceWrapperNode<Kokkos::Serial, Kokkos::HostSpace> > >::get (this=0x4010000000000000) at /home/xap/code/trilinos/fortrilinos/packages/teuchos/core/src/Teuchos_RCP.hpp:365
#3  0x00007ffff763c45d in _wrap_delete_TpetraMultiVector (farg1=0x7fffffffb9b0) at /home/xap/code/trilinos/fortrilinos/ForTrilinos/src/tpetra/src/swig/fortpetraFORTRAN_wrap.cxx:3391
#4  0x00007ffff7735aca in fortpetra::swigf_release_tpetramultivector (self=...) at /home/xap/code/trilinos/fortrilinos/ForTrilinos/src/tpetra/src/swig/fortpetra.F90:5439
#5  0x000055555555bbd9 in tpetramultivector_subviewnonconst (success=.TRUE.) at /home/xap/code/trilinos/fortrilinos/ForTrilinos/src/tpetra/test/test_tpetra_multivector.F90:723
#6  0x000055555555869f in test_tpetramultivector () at /home/xap/code/trilinos/fortrilinos/ForTrilinos/src/tpetra/test/test_tpetra_multivector.F90:46
#7  0x0000555555564100 in main (argc=1, argv=0x7fffffffc3fa) at /home/xap/code/trilinos/fortrilinos/ForTrilinos/src/utils/src/ForTrilinos.h:9
#8  0x00007ffff6653b97 in __libc_start_main (main=0x5555555640c0 <main>, argc=1, argv=0x7fffffffbee8, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffbed8) at ../csu/libc-start.c:310
#9  0x0000555555557c1a in _start ()
(gdb) 
```

aprokop:
OK, the bug was found and fixed.

PR #265, opened 2019-09-05
Initial setup for handling different nodes

MicheldeMessieres:
This PR is in progress and opened for discussion.

This will build for one node in the following priority:
Cuda -> OpenMP -> Serial

Tested on docker for OpenMP and Serial.
Tested on a linux GPU machine for Cuda.


aprokop:
ok to test

aprokop:
@MicheldeMessieres After thinking some more about the problem that we have, I think we may avoid carrying multiple copies. The fact is, the `.F90` files are the same (as they carry no knowledge about Kokkos at all), and the `.cxx` files are almost the same, except for the name of the node.

While it would be easy to implement if the `MAKE_SWIG_FORTRAN` did not do 2-in-1 (generate the wrappers *and* add the library), I think this can be avoided. We could use `ForTrilinos::DefaultNodeType` in the `fortpetra.i`. Then, we would include `ForTrilinos_DefaultNodeType.hpp`, which is configured _prior_ to calling SWIG. This way, the configuration of the node type is happening independent of the wrapper generation, which is what we want.

Here is a brief patch of what I'm thinking:
```diff
diff --git a/src/tpetra/src/CMakeLists.txt b/src/tpetra/src/CMakeLists.txt
index badc6ca..fd42c16 100644
--- a/src/tpetra/src/CMakeLists.txt
+++ b/src/tpetra/src/CMakeLists.txt
@@ -5,18 +5,18 @@
 INCLUDE_DIRECTORIES(${CMAKE_CURRENT_SOURCE_DIR})
 INCLUDE_DIRECTORIES(${CMAKE_CURRENT_BINARY_DIR})
 
-IF (ForTrilinos_ENABLE_DeveloperMode)
-  CONFIGURE_FILE(${CMAKE_CURRENT_SOURCE_DIR}/fortpetra.i.in
-    ${CMAKE_CURRENT_SOURCE_DIR}/${Kokkos_NODE_TYPE}/fortpetra.i @ONLY)
+CONFIGURE_FILE(${CMAKE_CURRENT_SOURCE_DIR}/ForTrilinos_DefaultNodeType.hpp.in
+  ${CMAKE_CURRENT_SOURCE_DIR}/ForTrilinos_DefaultNodeType.hpp @ONLY)
 
+IF (ForTrilinos_ENABLE_DeveloperMode)
   IF (ForTrilinos_EXPORT_SWIG)
     # Generate wrapper and proxy files since MPI is enabled
-    SET(CMAKE_SWIG_OUTDIR "${CMAKE_CURRENT_SOURCE_DIR}/swig/${Kokkos_NODE_TYPE}")
+    SET(CMAKE_SWIG_OUTDIR "${CMAKE_CURRENT_SOURCE_DIR}/swig")
   ENDIF()
 
   LIST(APPEND CMAKE_SWIG_FLAGS -I${CMAKE_CURRENT_SOURCE_DIR}/../../utils/src/)
   MAKE_SWIG_FORTRAN(MODULE fortpetra
-    SOURCE ${Kokkos_NODE_TYPE}/fortpetra.i
+    SOURCE fortpetra.i
     DEPLIBS ${Trilinos_LIBRARIES}
     EXTRASRC ${EXTRA_SOURCE}
     )
diff --git a/src/tpetra/src/fortpetra.i b/src/tpetra/src/fortpetra.i
index 530955d..e1526c1 100644
--- a/src/tpetra/src/fortpetra.i
+++ b/src/tpetra/src/fortpetra.i
@@ -21,21 +21,22 @@
 #define HAVE_TPETRA_INST_INT_LONG_LONG
 // From teuchos/kokkoscompat/src/KokkosCompat_ClassicNodeAPI_Wrapper.hpp
 
-%ignore @Kokkos_NODE_TYPE@;
-namespace Kokkos { namespace Compat {
-struct @Kokkos_NODE_TYPE@ {
+%ignore ForTrilinos::DefaultNodeType;
+namespace ForTrilinos {
+struct DefaultNodeType {
  static const bool classic = false;
 };
-} }
+}
 
 %{
 #include "Kokkos_DefaultNode.hpp"
+#include "ForTrilinos_DefaultNodeType.hpp"
 %}
 %inline %{
 typedef double                                  SC;
 typedef int                                     LO;
 typedef long long                               GO;
-typedef Kokkos::Compat::@Kokkos_NODE_TYPE@      NO;
+typedef ForTrilinos::DefaultNodeType      NO;
 typedef char                                    Packet;
 %}
```

Thoughts? Am I missing something? By the way, it also seems like this would avoid situations like in this PR, where `src/interface/src/swig` only has OpenMP and Serial, but no CUDA wrappers.

kruger:
This is a really good idea.     This also makes it easier to see how one could turn it into a runtime choice.  

BTW:  no CUDA wrappers?  I see them there.

aprokop:
@kruger There are CUDA wrappers for tpetra, but not for the interface (` src/interface/src/swig/KokkosCudaWrapperNode` is missing).

MicheldeMessieres:
@aprokop @kruger I'll update the missing Cuda interface wrappers - I thought I had committed them but seems I didn't.

aprokop:
@MicheldeMessieres Well, the new way would mean we won't have Serial/OpenMP/Cuda specific wrappers, just a generic one.

MicheldeMessieres:
@aprokop Ok I will just wait and update it all together with the new setup. Thanks.

MicheldeMessieres:
Updated the PR with the new ForTrilinos_DefaultNodeType.hpp setup,.

MicheldeMessieres:
Added the KOKKOS_ENABLE_CUDA just for the multivector tests so I could check it in the fortran file. I'm enabling the host/device tests only in that case. If we can include configs and get all the #defines in a more general way that would be better. I did not really think about it as I wasn't sure  if we wanted to organize cuda specific tests in this way.

aprokop:
@MicheldeMessieres Honestly, I was surprised to see that those `need_sync` and friends functions were exposed, I thought we hid them. But it actually makes sense now to fully expose them as they may be necessary for Cuda.

Are you familiar with the whole `sync` functionality in Tpetra classes? If one is using UVM, should a user call them?

MicheldeMessieres:
These tests are the first time I have worked with these sync calls directly. Since Tpetra currently requires UVM I'm not sure what implications the calls have for CUDA. I understand Tpetra would like to remove the UVM dependence eventually. I saw that things like need_sync_host() could be altered by calls to sync_host() so the API binding can be verified, despite being on UVM. That's as far as I went with these tests.

aprokop:
I got confirmation that we don't need any of the `sync` as of right now. So let's just get rid of that macro, and `%ignore` all the functions that depend on it; essentially, anything to do with host/device sync.

MicheldeMessieres:
I removed the sync tests and the macro.

aprokop:
@MicheldeMessieres I think you forgot to regenerate the wrappers.

MicheldeMessieres:
Updated the wrappers. Thanks.

aprokop:
Hmm, so the gcc builds with different node types pass. But the rebuilt container cannot run Flang for some reason, complaining about C++/Fortran compatibility. Not sure what's going on.

kruger:
I don't understand why it doesn't show up in our own containers, but since the error message says this: 
`relocation R_X86_64_32 against `.rodata' can not be used when making a PIE object; recompile with -fPIC`
perhaps adding this line:
`-DCMAKE_Fortran_FLAGS:STRING='-fPIC -pipe'` 
to the cmake commands will fix it.

aprokop:
Maybe, but I'd like to understand what is different. The container was changed to inherit from nvidia's docker container. It has newer compiler 7.4 by default, but that error can be reproduced even with 5.5. It worked before, and I don't see the difference right now.

aprokop:
While I'm trying to figure out clang, could someone rebase this PR on top of the current master to resolve conflicts?

aprokop:
If I install clang and clang++ in the container, the compatibility check passes.

kruger:
I had problems rebasing on top of master.  I think we need to squash the commits and then rebase because there are patches that undo each other.    Do you want me to do it, or do you?

MicheldeMessieres:
I can work on the rebas, but I'll have to do it tomorrow morning.

MicheldeMessieres:
Squashed, rebased to latest master, resolved minor conflict.

Andrey, I put your 3 commits combined into a single commit and then put all of mine into the 2nd commit.

aprokop:
retest this please

aprokop:
retest this please

MicheldeMessieres:
@aprokop I didn't have docker properly updated. I am building now and then I expect I'll see the failures.

aprokop:
Well, the situation is that it passes in the container when I run locally. But fails on Jenkins. Color me confused.

aprokop:
Oh, thinking about it, locally in the container I run a single build, while on Jenkins I run 3 *without* wiping out data. Let me try to wipe it out, and see what happens.

kruger:
I was getting weird build and configure errors with gcc until I completely blew away the build directory so this seems consistent.

aprokop:
One thing we should definitely do is to add a print out of the default node used by ForTrilinos by putting somewhere in cmake
```
MESSAGE(STATUS "ForTrilinos default node type: xxx")
```

aprokop:
retest this please

aprokop:
I think I found the issue. Originally, `ForTrilinos_DefaultNodeType.hpp` was generated in source. This meant that it is stored in Jenkins workspace. That one is not getting wiped out each time, and just checked out. It is likely that the untracked generated header got stuck there, and was being used instead of the one generated in the binary directory. So, despite selecting OpenMP node, it had Serial in the header. Lets see if tests pass now after I wiped out existing Jenkins workspace.

aprokop:
I can reproduce the test failures locally in the container, but not sure what's going on. It seems like it simply cannot create `TpetraMap`, for example, with CUDA node type. I'm going to try running without a container to see if tests fail.

MicheldeMessieres:
@aprokop I saw these errors:
cudaGetDeviceCount( & m_cudaDevCount ) error( cudaErrorInsufficientDriver): CUDA driver version is insufficient for CUDA runtime version 
Was not sure if that was a result of previous errors or causal.
I did not try Cuda docker yet but I did get my MPI non-docker running for Cuda. I need to make some system updates before I can install docker on this local machine it seems.

aprokop:
@#$@#$@ I completely missed that.

aprokop:
OK, I think the problem is that while container has CUDA, I have not properly set it up to use cuda runtime.

aprokop:
retest this please

aprokop:
retest this please

aprokop:
OK. The tests pass now, which is great. The CUDA build still takes an hour, but I think this can be brought down through ccache.

Issue #253, opened 2018-12-25
IoC drivers segfault with Flang when releasing base


aprokop:
I updated the container to Flang 20180612. The tests `ForTrilinosInterface_test_solver_handle_test_fortran` and `ForTrilinosInterface_test_nox_fortran` now segfault. Commenting out base class release fixes that.

I'm attaching a simpler [reproducer](https://github.com/trilinos/ForTrilinos/files/2708788/test_flang.txt). If base is commented out, it runs. Otherwise segfaults.

It is also funny that commenting out code inside `my_apply` while keeping base release also makes it pass. Weird.

aprokop:
For now, I workaround this issue in the drivers by hiding base release behind `__GNUC__`. I would have preferred to check for `__clang__` but that one seems undefined when compiling with Flang.

Issue #246, opened 2018-11-08
Fix missing RCPNode deallocations


aprokop:
In `Debug` mode, some tests in `Interface/test` spit out a bunch of warnings regarding dangling `RCPNode<blah>`. There are multiple unique sources of that. One of the known ones are:
- Something weird in setting up an implicit operator inside solver handle
- `SWIG_OWN` check in base class `release()` is never passed, it is always `SWIG_MOVE` instead

PR #268, opened 2019-10-07
Cleaning up RCP-related memory errors

jpopelar:
Addressing issues #246 and #253 

**A few notes regarding each:**
#246: There are still RCP errors present in `test_nox.F90` and `test_solver_handle.F90` specifically, though the amount of them is decreased, which should make further fixes easier to do

#253: The Flang segfaults mentioned in this report after uncommenting the suspected ``release()`` statements are no longer present. Considering this isn't an issue testing on Jenkins either, recommend closing this issue for now


kruger:
@jpopelar pointed out to me that for Issue #246 RCP errors are only showing up with gfortran/gcc and not with flang/clang.  Confirmed for me.  We could consider upgrading the gfortran in the docker image.

aprokop:
ok to test

PR #272, opened 2020-08-29
Redo testing from scratch using Jenkins pipelines

aprokop:


aprokop:
Merging as this will not pass current tests, but need to get `.jenkins` to `master`.

PR #271, opened 2020-08-11
Detribits (ForTrilinos as a standalone library)

sethrj:
This changes ForTrilinos to a standalone project that depends on Trilinos as a third-party library. Among other benefits, this will allow integration into system environments that already have Trilinos installed.

PR #274, opened 2020-08-29
Detribits part1

aprokop:
First part of splitting #271, rebased on master to launch new testing.


CMakeLists.txt
```diff
@@ -1,91 +1,184 @@
...
-
-TRIBITS_PROCESS_SUBPACKAGES()
-TRIBITS_PACKAGE_DEF()
-TRIBITS_PACKAGE_POSTPROCESS()
+#-----------------------------------------------------------------------------#
+# Test options
+#-----------------------------------------------------------------------------#
+
+option(ForTrilinos_TESTING "Build ForTrilinos unit tests" ON)
+option(ForTrilinos_EXAMPLES "Build ForTrilinos examples" ON)
```
aprokop, line 53: Should probably be `OFF`. Update testing if we decide to do that.

CMakeLists.txt
```diff
@@ -1,91 +1,184 @@
...
+#-----------------------------------------------------------------------------#
+
+find_package(Trilinos 12.17 REQUIRED MODULE
+  COMPONENTS
+    TeuchosCore TeuchosComm TeuchosNumerics TeuchosParameterList
+  OPTIONAL_COMPONENTS
+    # Optional
+    Belos
+    Tpetra
+    # Required for interface package
```
aprokop, line 94: Update comments

CMakeLists.txt
```diff
@@ -1,91 +1,184 @@
...
+set(ForTrilinos_USE_MPI "${Trilinos_USE_MPI}" CACHE BOOL
+  "MPI status depends on the Trilinos installation in use" FORCE)
+
+set(_fortrilinos_use_hl OFF)
+if(Trilinos_Anasazi_FOUND
+    AND Trilinos_NOX_FOUND
+    AND Trilinos_Stratimikos_FOUND
+    AND Trilinos_Thyra_FOUND
+    AND Trilinos_ThyraTpetraAdapters_FOUND)
+  set(_fortrilinos_use_hl ON)
```
aprokop, line 114: need to check if we need at least one of MueLu/Ifpack2

cmake/FindTrilinos.cmake
```diff
@@ -0,0 +1,94 @@
...
+    target_link_libraries("Trilinos::${_comp}"
+      INTERFACE Trilinos::Trilinos "${_lib}")
+  endforeach()
+  unset(_components)
+  unset(_comp)
+  unset(_lib)
+endif()
+
+
+if(Trilinos_Tpetra_FOUND AND PROJECT_NAME STREQUAL "ForTrilinos")
```
aprokop, line 63: Don't need the second check

scripts/build-yuri.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh -e
```
aprokop, line 1: Do we need to carry this file as part of ForTrilinos?

scripts/spack.yaml
```diff
@@ -0,0 +1,36 @@
+spack:
```
aprokop, line 1: Maybe rename this file to `spack_fortrilinos_env.yaml`

test/CMakeLists.txt
```diff
@@ -0,0 +1,164 @@
...
+    ${ARGN}
+  )
+
+  # Create executable
+  get_filename_component(_DIRNAME "${source}" DIRECTORY)
+  get_filename_component(_BASENAME "${source}" NAME_WE)
+  if(source MATCHES ".cpp$")
+    set(_BASENAME "${_BASENAME}/cpp")
+  endif()
+  string(REGEX REPLACE "[^a-zA-Z0-9_]+" "_" _TARGET "${_DIRNAME}/${_BASENAME}")
```
aprokop, line 67: Add a comment clarifying what this does.

aprokop:
I think I would actually prefer reverting the renaming of `fortrilinos` to `fortrilinos_hl` so as not to spook current users.

aprokop:
`cathode` ran out of space, so I can't really test this right now. Need to wait till Monday.

CMakeLists.txt
```diff
@@ -1,91 +1,184 @@
...
-  SET(CMAKE_Fortran_FLAGS "-ffree-line-length-none ${CMAKE_Fortran_FLAGS}")
+#---------------------------------*-CMake-*----------------------------------#
+# Copyright 2020 UT-Battelle, LLC
+# License-Filename: LICENSE
+# SPDX-License-Identifier: BSD-3-Clause
+#----------------------------------------------------------------------------#
+
+cmake_minimum_required(VERSION 3.12)
+project(ForTrilinos VERSION 2.0.0 LANGUAGES Fortran CXX)
+cmake_policy(VERSION 3.12...3.18)
```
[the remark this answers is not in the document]
sethrj: This just sets the default policy to "OLD" for policies introduced after 3.18. So if for example a new UseSWIG policy is made in 3.20, we wouldn't have to explicitly set it to OLD to avoid a warning.

CMakeLists.txt
sethrj: OK. Both, or just examples?

cmake/FindTrilinos.cmake
sethrj: This script is installed alongside `ForTrilinosConfig.cmake` and is used to find the original version of Trilinos. When another project does `find_package(ForTrilinos)`, the logic there prevents that downstream project from rerunning the checks.

scripts/build-yuri.sh
sethrj: No, but I find it useful when a software package has a couple of build scripts so that you can see an expected use case. In the "master" version of fortrilinos now, it's difficult to guess what the right cmake flags are to get fortrilinos compiling as expected. I can delete the file if you find it distracting.

sethrj:
> I think I would actually prefer reverting the renaming of `fortrilinos` to `fortrilinos_hl` so as not to spook current users.

The problem is this would propagate two CMake targets: `ForTrilinos::fortrilinos`  for the module itself and `ForTrilinos::ForTrilinos` for the cmake target.

(Also, who are the current users? Maybe we could discuss the naming with them.)

CMakeLists.txt
aprokop: Both.

CMakeLists.txt
aprokop: Just leave it as is for now. If it breaks, we'll fix it .

cmake/FindTrilinos.cmake
aprokop: Ah, OK. This reminds me. It may make sense to store the found Trilinos version in the installation, so that the users won't need to know about Trilinos when using ForTrilinos. There are pros and cons, but overall it may be worth it. We do that in [ArborX](https://github.com/arborx/ArborX/blob/4834bff44c23c9510c6ed93366638dcdf85ab217/cmake/ArborXConfig.cmake.in#L7-L12) for Kokkos.

scripts/build-yuri.sh
aprokop: I'm OK with it. These just tend to procreate when giving perfect conditions of being tucked in little used directory.

aprokop:
retest this please

cmake/FindTrilinos.cmake
sethrj: Yes, the required trilinos directory gets exported alongside the ForTrilinos cmake so it will get automatically located -- I think this is what you're asking.

scripts/spack.yaml
sethrj: I'm just going to delete the environment, since I'm going to create a spack package once this PR goes through.

PR #273, opened 2020-08-29
Detribits

aprokop:
First part of splitting #271, rebased on master to launch new testing.

PR #275, opened 2020-08-31
Update to Trilinos 12.18.1

aprokop:
Picking second half of #273.


aprokop:
It passes the tests, but has few warnings. It can be merged. Will need to figure out what we want to do with warnings, especially those coming from Trilinos.

PR #280, opened 2020-09-18
Update ForTrilinos for SWIG 4.0.2+fortran release

sethrj:
Update SWIG interface files for changes upstream, and regenerate with the SWIG release v4.0.2+fortran.


aprokop (commented):
Yeah, there is no way I'm doing a proper review on this. I briefly browsed the first commit, and what I saw looked ok.

sethrj:
The SWIG interface file changes are pretty limited (but as usual with SWIG took hours to figure out),  but the fortran interfaces are essentially unchanged.  The only real interface change is that now "native" fortran character arrays are used by default. That's no change for GCC-compiled versions, but the `xlf` compiler does distinguish between `kind=c_char` and native characters, so it'll be easier to compile application code for that system.

aprokop:
This sounds good to me.

PR #281, opened 2020-09-18
Update wrappers for Trilinos v13.0

sethrj:
A single minor change was needed for the trilinos interface. I've added documentation for the SWIG and Trilinos version compatibilities.

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -9,6 +9,34 @@ its only dependency. ForTrilinos can be installed through `the Spack HPC
 package manager <https://spack.readthedocs.io/en/latest/>`_ or
 independently from your local installation of Trilinos.
 
+.. _version:
+
+Version compatibility
+---------------------
+
+ForTrilinos wrappers are tightly coupled to the Trilinos API, so upstream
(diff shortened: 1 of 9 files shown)


doc/install.rst
```diff
@@ -9,6 +9,34 @@ its only dependency. ForTrilinos can be installed through `the Spack HPC
...
+.. table:: Version compatibility table for ForTrilinos.
+
+   ===========  ============== ======================
+   ForTrilinos  Trilinos       SWIG
+   ===========  ============== ======================
+   2.0.0        13.0           4.0.2+fortran
+   2.0.0-dev3   12.18.1        4.0.2+fortran
+   2.0.0-dev2   12.18.1        4.0.0+fortran+15e6ed59
+   2.0.0-dev1   12.17+8a82b322 4.0.0+fortran+15e6ed59
+   1.0          12.8.1         ---
```
aprokop, line 33: Maybe explain that `+fortran` is coming from spack.

aprokop (approved):

doc/install.rst
sethrj: Added a paragraph explaining the unusual versioning.

PR #282, opened 2020-09-25
Final cleanup for 2.0.0 release

sethrj:
- Add verbosity to the anasazi output to debug it
- Don't fail when solver doesn't converge -- eigenvalues may be right but just to a lower tolerance
- Add accessors to determine convergence and iteration count
- Re-enable formerly failing tests and add additional tetsing
- Remove zombie files


sethrj:
I can do a clang-format on the source code in another PR just to keep things clean.

aprokop:
retest this please

src/fortrilinos_hl/eigen_handle.hpp
```diff
@@ -44,33 +44,30 @@ namespace ForTrilinos {
...
 
     // Setup operators
-    void setup_operator    (const Teuchos::RCP<Operator>& A);
+    void setup_operator(const Teuchos::RCP<Operator>& A);
     void setup_operator_rhs(const Teuchos::RCP<Operator>& M);
 
     // Setup solver based on the parameter list
     void setup_solver(const Teuchos::RCP<Teuchos::ParameterList>& paramList);
+    // Maximum number of eigenvalues to solve for
+    int max_eigenvalues() const;
```
aprokop, line 61: What is the use case for this function? Also, the comment seems to indicate as if its purpose is to specify that value, instead of getting it out?

src/fortrilinos_hl/eigen_handle.cpp
```diff
@@ -126,24 +120,35 @@ namespace ForTrilinos {
     solver_ = Anasazi::Factory::create(solverName, problem, paramList->sublist(solverName));
 
     status_ = SOLVER_SETUP;
+    paramList_ = paramList;
```
aprokop, line 123: Why would we want to store it? In fact, I'm not sure why there is `paramList_` in the first place.

src/fortrilinos_hl/eigen_handle.cpp
```diff
@@ -126,24 +120,35 @@ namespace ForTrilinos {
...
+
+  int TrilinosEigenSolver::max_eigenvalues() const {
+    TEUCHOS_ASSERT(status_ >= SOLVER_SETUP);
+    return numEigenvalues_;
   }
 
   int TrilinosEigenSolver::solve(Teuchos::ArrayView<SC> eigenValues,
                                  Teuchos::RCP<MultiVector>& eigenVectors,
-                                 Teuchos::ArrayView<int> eigenIndex) const {
+                                 Teuchos::ArrayView<int> eigenIndex) {
```
aprokop, line 133: Not a fan, but I see why you need it. Could also make `converged_` and friends `mutable`.

src/fortrilinos_hl/eigen_handle.cpp
```diff
@@ -126,24 +120,35 @@ namespace ForTrilinos {
...
+    TEUCHOS_ASSERT(status_ >= SOLVER_SETUP);
 
     Anasazi::ReturnType r = solver_->solve();
-    TEUCHOS_ASSERT(r == 0);
+    status_ = SOLVED;
+    converged_ = (r == Anasazi::Converged);
+    numIters_ = solver_->getNumIters();
+    if (!converged_ && comm_->getRank() == 0) {
+      std::cout << "fortrilinos: warning: anasazi solver failed to converge after "
+          << numIters_ << " iterations" << std::endl;
```
aprokop, line 145: This may not follow xSDK or E4S policies about not hardcoding output.

src/fortrilinos_hl/eigen_handle.cpp
```diff
@@ -163,18 +168,16 @@ namespace ForTrilinos {
 
     eigenVectors = solution.Evecs;
 
-    return eValues.size();
+    return numConverged;
```
aprokop, line 171: This changes the meaning. What do we actually want eigensolver to return? Number of converged values, or the size of the array with eigenvalues? The issue is that if it's the former, there is no easy way to get the size of the latter.

aprokop (changes requested):
Solid work. I have just a few minor comments and questions.

ROADMAP.md
```diff
@@ -1,858 +0,0 @@
-# ForTrilinos roadmap
```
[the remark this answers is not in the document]
sethrj: As you say, it was a pain to maintain and I don't think it *was* being maintained; I think it was a relic of when we first started. I didn't think it provided anything particularly useful -- I think our time would be better spent trying to autodocument the fortran interface.

src/fortrilinos_hl/solver_handle.hpp
```diff
@@ -43,14 +43,8 @@ namespace ForTrilinos {
...
-    TrilinosSolver() : status_(NOT_INITIALIZED) { }
-
-    TrilinosSolver(const TrilinosSolver&) = delete;
-    void operator=(const TrilinosSolver&) = delete;
-
-    // Initialize
-    void init();
-    void init(const Teuchos::RCP<const Teuchos::Comm<int>>& comm);
+    TrilinosSolver();
+    explicit TrilinosSolver(const Teuchos::RCP<const Teuchos::Comm<int>>& comm);
```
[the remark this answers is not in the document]
sethrj: I generally prefer to keep implementation details out of the main header files (the DefaultComm header would have to move to the `.hpp`), and I also generally prefer not to use default arguments in C++.

src/fortrilinos_hl/solver_handle.hpp
```diff
@@ -43,14 +43,8 @@ namespace ForTrilinos {
   public:
 
     // Constructors
-    TrilinosSolver() : status_(NOT_INITIALIZED) { }
-
-    TrilinosSolver(const TrilinosSolver&) = delete;
-    void operator=(const TrilinosSolver&) = delete;
```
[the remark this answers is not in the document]
sethrj: These methods were ignored by SWIG anyway (none of the assignment/copying logic changed in the wrapper code) and there's nothing mechanically wrong with copying the classes -- so I figured the deleted methods didn't need to be here.

test/fortrilinos_hl/data/davidson.xml
```diff
@@ -6,9 +6,11 @@
     <Parameter name="Maximum Subspace Dimension" type="int" value="25"/>
     <Parameter name="Restart Dimension" type="int" value="5"/>
     <Parameter name="Maximum Restarts" type="int" value="20"/>
+    <Parameter name="Verbosity" type="int" value="9"/>
   </ParameterList>
   <Parameter name="Preconditioner Type" type="string" value="MueLu"/>
   <ParameterList name="MueLu">
-    <Parameter name="verbosity" type="string" value="none"/>
+    <Parameter name="verbosity" type="string" value="high"/>
+    <Parameter name="debug: graph level" type="int" value="4"/>
```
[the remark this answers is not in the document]
sethrj: Haha good point -- I was digging around trying to figure out what was going on with the convergence.

src/fortrilinos_hl/eigen_handle.hpp
```diff
@@ -44,33 +44,30 @@ namespace ForTrilinos {
...
-    TrilinosEigenSolver() : status_(NOT_INITIALIZED) { }
-
-    TrilinosEigenSolver(const TrilinosEigenSolver&) = delete;
-    void operator=(const TrilinosEigenSolver&) = delete;
-
-    // Initialize
-    void init();
-    void init(const Teuchos::RCP<const Teuchos::Comm<int>>& comm);
+    TrilinosEigenSolver();
+    explicit TrilinosEigenSolver(const Teuchos::RCP<const Teuchos::Comm<int>>& comm);
```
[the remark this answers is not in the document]
sethrj: I guess it's a style preference, but I think having a default constructor and an explicit one-argument constructor is more clear than a combined constructor with a default argument. The SWIG code it generates is the same, anyway.

src/fortrilinos_hl/eigen_handle.hpp
sethrj: To extract the results when you call `solve`, the arrays you pass from fortran have to be sized correctly in advance. This function is the way to get that value. I could see the comment being ambiguous; do you have a suggestion for an improved phrasing?

src/fortrilinos_hl/eigen_handle.cpp
sethrj: I'm not sure... the `paramList_` member data was in there originally and left unused. I'll delete it.

src/fortrilinos_hl/eigen_handle.cpp
sethrj: Since `solve` is changing the state of the managed `solver_` object regardless of the `converged_` status update, I think it's more semantically correct for the member function to be non-const.

src/fortrilinos_hl/eigen_handle.cpp
sethrj: Ooh, good point. There's also a `cout` for the solve result in `solver_handle`. Should I just remove the cout in both places?

src/fortrilinos_hl/eigen_handle.cpp
sethrj: I remember puzzling over this as well. I guess I assumed that unconverged eigenvalues would not be useful? And I thought the size of the array would be the same as `max_eigenvalues`.

sethrj:
@aprokop Thanks for the comments! I'm going to wait for your responses before pushing new changes.

ROADMAP.md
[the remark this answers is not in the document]
aprokop: One of the things it provided was whether there is a performance penalty for using a particular function, as some of them required copying arrays and doing +-1. However, as I personally am unlikely to spend time maintaining this, I cannot fault you for removing it.

src/fortrilinos_hl/solver_handle.hpp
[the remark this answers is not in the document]
aprokop: > I also generally prefer not to use default arguments in C++.

nani?! why?

src/fortrilinos_hl/solver_handle.hpp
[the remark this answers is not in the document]
aprokop: My concern is that we don't make it clear the handle semantics. I.e., if you copy a handle, and modify its matrices, is that going affect the other copy? As of right now, it will, and that's not clear. I didn't realize that SWIG is ignoring them.

src/fortrilinos_hl/eigen_handle.hpp
aprokop: My concern is with complex eigenvalues. If you look at how anasazi does them, the size of the arrays will depend on whether the found eigenvalue is real or complex. So, the number of eigenvalues by itself is not sufficient to make arrays of the right size (though, you *could* use it to provide an upper bound by multiplying it by 2). 

But my overall thinking is that a user is the one who constructs the parameter list to solve the eigenproblem. Thus, he has a knowledge of how many eigenvalues he/she needs and can use that to size the arrays.

src/fortrilinos_hl/eigen_handle.cpp
aprokop: Either remove, or provide yet another option to control verbosity. I assume that this information can be gleaned after the call to `solve()`, if we add a function to get out the number of iterations.

src/fortrilinos_hl/solver_handle.hpp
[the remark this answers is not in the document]
sethrj: Mostly because they can get out of hand too quickly: I've seen constructors with like 5 optional int/double arguments (overloaded with other constructors as well) and when you see the code that calls the constructor it is impossible to tell what it's calling.

Having an indpendent empty constructor also makes the code more readable -- if I only saw the one-argument constructor I might not make it all the way to finding the default argument, especially when the type is so complex and the default value even more so.

Finally, the most concrete reason is that `explicit` doesn't work correctly with one-argument default constructors: see for example https://wg21.link/P0935R0 with changes to the default library.

src/fortrilinos_hl/solver_handle.hpp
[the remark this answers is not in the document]
sethrj: That's true... the ever-present danger of shared pointers.

src/fortrilinos_hl/eigen_handle.cpp
aprokop: If I remember this right, there is a second array of indices. 0 means it's a real eigenvalue, while 1 and -1 mean complex eigenvalue and its conjugate, respectively. So, you can get three eigenvalues, one real and 2 complex. In that situation, I don't recall what the `eValues` is going to be sized to.

src/fortrilinos_hl/eigen_handle.hpp
sethrj: Oh really, I didn't realize that the array size depended on whether they were real. I knew that we required a factor of two space on the incoming vector for real/imaginary.

From a non-regular-trilinos-user perspective it wasn't immediately obvious to me (looking at existing test code) how to get the eigenvalues. And I really really dislike parameter lists as the primary means of communication between libraries: they're extremely fragile (consider the arbitrary capitalization and spelling of `NumEV`) and you literally have to dig into the source code (rather than looking at a class interface) to figure out what the names and types of parameters are. 😬

src/fortrilinos_hl/eigen_handle.cpp
sethrj: I've removed it -- it'll be up to the caller to check and make sure the values are converged.

src/fortrilinos_hl/solver_handle.cpp
```diff
@@ -104,18 +95,7 @@ namespace ForTrilinos {
...
-  void TrilinosSolver::finalize() {
-    // No need to check the status_, we can finalize() at any moment.
-    comm_          = Teuchos::null;
-    A_             = Teuchos::null;
-    paramList_     = Teuchos::null;
-    thyraInverseA_ = Teuchos::null;
-
-    status_ = NOT_INITIALIZED;
+    if (map->getComm()->getRank() == 0)
+      std::cout << "solve result:" <<  status << std::endl;
```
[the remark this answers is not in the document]
sethrj: Deleted this `cout` per our other discussion

src/fortrilinos_hl/eigen_handle.hpp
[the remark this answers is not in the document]
sethrj: Actually I remembered the `explicit` correctness issue as I posted on the other constructor discussion --  closing.

src/fortrilinos_hl/fortrilinos_hl.i
```diff
@@ -30,6 +30,7 @@ typedef char                                    Packet;
 #include "fortrilinos_hl/eigen_handle.hpp"
 %}
 
+%fortransubroutine ForTrilinos::TrilinosSolver::solve;
```
aprokop, line 33: What does this do?

src/fortrilinos_hl/solver_handle.cpp
```diff
@@ -81,7 +81,7 @@ TrilinosSolver::TrilinosSolver(const Teuchos::RCP<const Teuchos::Comm<int>>& com
     status_ = SOLVER_SETUP;
   }
 
-  void TrilinosSolver::solve(const Teuchos::RCP<const MultiVector>& B, Teuchos::RCP<MultiVector>& X) const {
+  double TrilinosSolver::solve(const Teuchos::RCP<const MultiVector>& B, Teuchos::RCP<MultiVector>& X) const {
```
aprokop, line 84: Honestly, this is a weird signature of the function. I don't think it should return tolerance. The only thing user really cares about is whether it converged or not.

src/fortrilinos_hl/fortrilinos_hl.i
sethrj: Instead of being a function (reqiuring `tol = foo%solve(...)`) it changes the signature so that `solve` is still a subroutine, but it returns an optional value through an argument `foo%solve(..., tol)`. This lets us return the solution tolerance without requiring changes in existing code, and it makes more sense for solve to be a subroutine since it returns the solution as an argument anyway.

src/fortrilinos_hl/solver_handle.cpp
sethrj: OK, so we could make it return a bool instead. I was just trying to return the single scalar with the most information about the solve.

sethrj:
OK @aprokop , any more feedback? Were the earlier discussions resolved to your satisfaction?

aprokop (approved):

PR #288, opened 2022-06-03
Update based on minor SWIG changes

sethrj:
The only substantial change is that `size_t` is no longer equal to a `long` (which it never should have been), so this is a bug fix. On most 64-bit platforms, I think `c_size_t` and `c_long_long` should be equivalent: so this should not result in any downstream changes.


sethrj:
@aprokop The CI [seems to be failing](https://cloud.cees.ornl.gov/jenkins-ci/job/ForTrilinos2/10/execution/node/18/log) because the CUDA docker image has apparently been deleted due to its age. Do you have time to upgrade the docker image?

aprokop:
Seems like CI passed, but the status wasn't reported back.

PR #291, opened 2023-01-19
Support ForTrilinos 13.4

sethrj:
The main change is to use a newer SWIG version (no substantial changes) and to delete functions that were deprecated in Tpetra. We *have* to build without deprecated functions because the pre-generated code will result in compile errors if linked against a version of trilinos built with `Tpetra_ENABLE_DEPRECATED_CODE`. 

Not only does the compiler switch result in completely different APIs for the same version of a code, but some interfaces were *removed* in the 13.x series: trilinos/Trilinos#9865 for example marked `StaticProfile` as deprecated, and trilinos/Trilinos#10552 deleted it! There was no Trilinos release in between them. This completely breaks forward and backward compatibility even with deprecated code enabled.


sethrj:
Ugh I didn't have the unit tests or examples built. And it looks like some interfaces were *removed* in the 13.x series: #9865 for example marked `StaticProfile` as deprecated, and #10552 deleted it! There was no Trilinos release in between them. This completely breaks forward and backward compatibility even with deprecated code enabled.

aprokop:
> Not only does the compiler switch result in completely different APIs for the same version of a code, but some interfaces were _removed_ in the 13.x series: [trilinos/Trilinos#9865](https://github.com/trilinos/Trilinos/pull/9865) for example marked `StaticProfile` as deprecated, and [trilinos/Trilinos#10552](https://github.com/trilinos/Trilinos/pull/10552) deleted it! There was no Trilinos release in between them. This completely breaks forward and backward compatibility even with deprecated code enabled.

Wow. That's messed up.

